PR #67, opened 2023-09-07
feat/fix: Updated passkey Implementation for supabase

tipusinghaw:
### What ###
- Stored password while adding the passkey
- Changed login, and signup  approach for passkey
     - Login Approach with Passkey:
        Users can now log in using their passkeys in addition to their regular usernames or email addresses.
        The application has likely added a feature to authenticate users based on their passkey, which might involve decrypting or verifying the passkey to grant access.

    - Signup Approach with Passkey:
        During the signup process, users may have the option to create a passkey in addition to a traditional password.
        The application may enforce certain security requirements for passkeys, such as length and complexity.


nishad-ayanworks (approved):
LG

PR #87, opened 2023-09-15
refactor: implemented org and user to by default public

nishad-ayanworks:
-  Implemented org and user to by default public
- Checked the same org name is exist in platform for update organization


tipusinghaw (approved):
LGTM

PR #105, opened 2023-10-02
feat: ecosystem ms setup

tipusinghaw:
### What ###
- Implemented ecosystem microservice


nishad-ayanworks (approved):
LG

libs/prisma-service/prisma/migrations/20231002114401_ecosystem_setup/migration.sql
```diff
@@ -0,0 +1,90 @@
+-- CreateTable
+CREATE TABLE "ecosystem_roles" (
+    "id" SERIAL NOT NULL,
```
ajile-in, line 3: As agreed, let's use UUID instead of an incrementing number

PR #119, opened 2023-10-05
refactor: register user details

tipusinghaw:
### What ###
- Updated `user` DTO for first name and last name


nishad-ayanworks (approved):
LG

PR #122, opened 2023-10-05
fix: resolved build error for prisma client

bhavanakarwade:
## What ## 
- Resolved build error for Prisma Client.


nishad-ayanworks (approved):
LG

PR #128, opened 2023-10-07
feat: delete pending ecosystem invitations

bhavanakarwade:
## What ## 
- Implemented delete ecosystem pending invitations API.

## Why ##
- By removing obsolete pending invitations, we ensure that only valid and relevant data exists, enhancing both user experience and system performance.


nishad-ayanworks (approved):
LG

PR #131, opened 2023-10-07
feat: get endorsement transactions

nishad-ayanworks:
## What ##
- Implemented GET API for endorsement transactions with details of schema and credential definition

## Why ##


libs/prisma-service/prisma/data/credebl-master-table.json
```diff
@@ -107,5 +107,43 @@
       "registerDIDPayload": "",
       "indyNamespace": "indicio:testnet"
     }
+  ],
```
tipusinghaw, line 110: Please these dummy values

tipusinghaw (approved):
LGTM

PR #132, opened 2023-10-09
fix: endorsement transactions

nishad-ayanworks:
## What ##
- Fixed GET API for endorsement transactions with details of schema and credential definition

## Why ##


KulkarniShashank (approved):
LGTM

PR #134, opened 2023-10-09
fix: endorsor transaction list 

nishad-ayanworks:
## What ##
- Removed redundant function of endorser transaction list

## Why ##


KulkarniShashank (approved):
LGTM

PR #148, opened 2023-10-12
refactor: ecosystem config settings

nishad-ayanworks:
## What ##
- Updated `ecosystem_orgs` model for `deploymentMode' parameter
- Refactored `ecosystem_config` to set all configurations with `key` and `value`.

## Why ##
- To configure the ecosystem configurations


tipusinghaw (approved):
LGTM

PR #149, opened 2023-10-12
fix: Added validation for schema and cred-def and updated payload for create request cred-def

KulkarniShashank:
### What ###
Added validation for the schema and credential definition (cred-def).
Changed the payload to include the value and type for the create request credential definition (cred-def).

### Why ###
The changes were made to enhance the security and accuracy of the credential creation process. By implementing schema and cred-def validation, we can ensure that the data being used is correct and secure.

### How ###
Implemented schema and cred-def validation in the code.
Updated the payload structure to include "value" and "type" for create request credential definitions.


tipusinghaw (approved):
LG

tipusinghaw (approved):

PR #150, opened 2023-10-12
refactor: removed org validation for reject

tipusinghaw:
### What ###
- Removed `orgId` invalid validation for reject endorsement request


nishad-ayanworks (approved):
LG

PR #153, opened 2023-10-13
feat: Schema and cred-def auto sign and submit endorsement transaction 

KulkarniShashank:
### What ###

- Enables the automatic signing and submission of endorsement transactions, incorporating schema and credential definition information.

### Why ###

- This feature will simplify the process of endorsing transactions by automating the signing and submission steps. It will reduce the manual effort required from users, making the system more user-friendly.
- Automation can significantly speed up the transaction endorsement process, leading to quicker turnaround times and improved efficiency for users.

### How ###

- To achieve the automatic signing and submission of endorsement transactions with schema and credential definitions
- The system will fetch the relevant schema and credential definition data from the appropriate sources or repositories.
- The retrieved data will be integrated into the transaction, ensuring that the endorsement process includes the latest schema and credential information.
- The system will automatically sign the transaction using the user's private key or the designated signing mechanism.
- After signing, the system will automatically submit the endorsed transaction to the network


tipusinghaw (approved):
LGTM

PR #176, opened 2023-10-18
fix: added attributes as array of string

tipusinghaw:
#### What ####
- added attributes as an array of strings for request credential-definition


KulkarniShashank (approved):
LG

PR #179, opened 2023-10-19
fix: send invitation flow for organization & ecosystem

nishad-ayanworks:
## What ##
- Restricted self invitation for organization & ecosystem join
- Allowed to send invitation again if user has rejected

## Why ##
- To restrict the user self invitation
- To able to send invitation again if rejected unconditionally


apps/ecosystem/src/ecosystem.controller.ts
```diff
@@ -97,9 +97,9 @@ export class EcosystemController {
    */
   @MessagePattern({ cmd: 'send-ecosystem-invitation' })
   async createInvitation(
-    @Body() payload: { bulkInvitationDto: BulkSendInvitationDto; userId: string }
+    @Body() payload: { bulkInvitationDto: BulkSendInvitationDto; userId: string, userEmail: string }
```
tipusinghaw, line 100: Please remove the `body` decorator

nishad-ayanworks: Resolved

PR #192, opened 2023-10-25
feat: bulk-issuance APIs

tipusinghaw:
### What ### 
- Implemented download CSV by credential definition.
- Implemented get all list of credential definitions mapped with schema.

### DIFF
--- a/apps/api-gateway/src/issuance/issuance.service.ts
+++ b/apps/api-gateway/src/issuance/issuance.service.ts
@@ -3,6 +3,7 @@ import { ClientProxy } from '@nestjs/microservices';
 import { BaseService } from 'libs/service/base.service';
 import { IUserRequest } from '@credebl/user-request/user-request.interface';
 import { IssuanceDto, IssueCredentialDto } from './dtos/issuance.dto';
+import { FileExportResponse } from './interfaces';
 
 @Injectable()
 export class IssuanceService extends BaseService {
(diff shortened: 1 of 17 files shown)


apps/api-gateway/src/issuance/issuance.service.ts
```diff
@@ -50,4 +51,10 @@ export class IssuanceService extends BaseService {
         return this.sendNats(this.issuanceProxy, 'webhook-get-issue-credential', payload);
     }
 
+    async exportSchemaToCSV(credentialDefinitionId: string
+        ): Promise<FileExportResponse> {
+          const payload = {credentialDefinitionId};
+          return (await this.sendNats(this.issuanceProxy, 'export-schema-to-csv-by-credDefId', payload)).response;
```
KulkarniShashank, line 57: there's no need to add await if you return the function directly.

apps/issuance/src/issuance.service.ts
```diff
@@ -313,4 +316,54 @@ export class IssuanceService {
...
+        return Promise.reject('Unable to transform schema data for CSV.');
+      }
+
+      const csv = parse(csvFields, { fields: csvFields });
+
+      const filePath = join(process.cwd(), `uploadedFiles/exports`);
+
+
+      let processedFileName: string = credentialDefinitionId;
+      processedFileName = processedFileName.replace(/[\/:*?"<>|]/g, '_');
```
KulkarniShashank, line 344: can we direct use this login on credentialDefinitionId processedFileName.replace(/[\/:*?"<>|]/g, '_')

KulkarniShashank (approved):
LG

PR #195, opened 2023-10-27
fix: get all schema and credentail definition for bulk operation

tipusinghaw:
#### What ####
- fix the tag issue for schema and credential-definition


KulkarniShashank (approved):
LGTM

PR #210, opened 2023-10-30
fix: orgId parameter for ecosystem schema

tipusinghaw:
### What ###
- Added orgId in params to get all ecosystem schemas


nishad-ayanworks (approved):
LG

PR #220, opened 2023-11-01
fix: refactor connection handling in the connection service

KulkarniShashank:
### What ###
This connection invitation related issues within the connection service.

### Why ###
The primary goal is to improve the handling of connection invitations, ensuring that a new invitation is not generated if one already exists. This will enhance the overall efficiency and reliability of the service.

### How ###
Identified the issue related to connection invitations.
Implemented a check to prevent the generation of a new invitation when one already exists.
Conducted thorough testing to verify the effectiveness of the fix.

### DIFF
--- a/apps/connection/src/connection.repository.ts
+++ b/apps/connection/src/connection.repository.ts
@@ -64,6 +64,28 @@ export class ConnectionRepository {
     }
 
     /**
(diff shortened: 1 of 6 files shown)


nishad-ayanworks (approved):
LGTM

PR #233, opened 2023-11-07
feat: get uploaded file details and file data APIs

tipusinghaw:
#### What ####
- Implemented API to get all uploaded files for bulk operation
- Implemented API to get file details based upon `fileId` for bulk operation


nishad-ayanworks (approved):
LG

PR #250, opened 2023-11-09
feat: added S3 bucket for file upload

tipusinghaw:
#### What ####
- Setup S3 bucket in libs
- Implemented file upload logic for bulk-issuance


KulkarniShashank (approved):
LG

PR #251, opened 2023-11-10
fix: connection pool of prisma postgres

nishad-ayanworks:
## What ##
-  Included pool database URL in schema prisma

## Why ##
- To use the connection pooling for postgres DB.


KulkarniShashank (approved):
LGTM

PR #255, opened 2023-11-10
Disallowed dummy email domains

pallavicoder:
### What ###

- Disallowed email from disposable-email-domains

### How ###

- By adding function to check  disposable-email-domains


nishad-ayanworks (approved):
LG

PR #260, opened 2023-11-15
fix:modified dtos class validators messages

pallavicoder:
### What ###

- Modified dto's class validator messages

### Why ###

- To fix a bug

###  How ###

- By modifying class validators messages


tipusinghaw (approved):
LGTM

PR #274, opened 2023-11-21
fix: added label for the out-of-band issuance and verification

KulkarniShashank:
1. Added the label on the out-of-band issuance and verification payload


nishad-ayanworks (approved):
LG

PR #289, opened 2023-11-22
fix: file upload blob issue

tipusinghaw:
#### What ####
- Added name as string
- Added `partially_completed` status


nishad-ayanworks (approved):
LG

PR #290, opened 2023-11-22
Error message emitted during the bulk issuance process failed through sockets

pallavicoder:
### What ###

-  Emitting error message  during the bulk issuance process failed  through socket

### Why ###

- To inform the end user that the bulk issuance process has encountered an error.


tipusinghaw (approved):
LGTM

PR #301, opened 2023-11-24
feat: network url fetch on the ledger table

KulkarniShashank:
### What ###
 Introducing network URL fetch functionality on the ledger table.

### Why ###
 To enable the ledger table to fetch data from external URLs, providing a more dynamic and versatile way of populating the ledger

### How ###
 Implemented a function that accepts a URL as a parameter, retrieves data from the URL, and updates the ledger table with the fetched data.


nishad-ayanworks (approved):
LG

PR #306, opened 2023-11-25
refactor: connection id updation in oob

nishad-ayanworks:
## What ##
-  Updated connection id updation for oob in table


KulkarniShashank (approved):
LG

PR #319, opened 2023-11-29
refactor: map ecosystem organization

nishad-ayanworks:
## What ##
- Refactored migration file of ecosystem org mapping for orgId column.


KulkarniShashank (approved):
LG

PR #320, opened 2023-12-01
feat: oob issuance for multiple records

bhavanakarwade:
## What ## 
- Refactor POST API for Out-of-band issuance for multiple records.

## Why ##
- To create multiple credential offer at same time.
- To handle and process multiple records at same time.


KulkarniShashank (approved):
LG

PR #333, opened 2023-12-06
fix: Validations in endorsement requests.

nishad-ayanworks:
## What ##
-  Included validations for the payload of endorsement requests.


KulkarniShashank (approved):
LG

PR #337, opened 2023-12-07
Afj spinup script change

DeepakNemad:
## What?
- Changes in AFJ spinup script, attach new EFS volume and mount point.

## Why?
- To avoid SSH access of servers


Sheetal-ayanworks:
resolved

PR #338, opened 2023-12-07
fix: Applied trim validations 

bhavanakarwade:
## What ## 
- Applied trim validations.

### DIFF
--- a/apps/api-gateway/src/issuance/dtos/issuance.dto.ts
+++ b/apps/api-gateway/src/issuance/dtos/issuance.dto.ts
@@ -135,7 +135,7 @@ export class OutOfBandCredentialDto {
     @IsEmail()
     @IsNotEmpty({ message: 'Please provide valid email' })
     @IsString({ message: 'email should be string' })
-    @Transform(({ value }) => value.trim())
+    @Transform(({ value }) => value.trim().toLowerCase())
     @IsOptional()
     emailId: string;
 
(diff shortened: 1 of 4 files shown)


apps/api-gateway/src/issuance/dtos/issuance.dto.ts
```diff
@@ -1,18 +1,36 @@
-import { IsArray, IsNotEmpty, IsOptional, IsString, IsEmail, ArrayMaxSize } from 'class-validator';
+import { IsArray, IsNotEmpty, IsOptional, IsString, IsEmail, ArrayMaxSize, ValidateNested } from 'class-validator';
 import { ApiProperty } from '@nestjs/swagger';
 import { Transform, Type } from 'class-transformer';
 import { toNumber } from '@credebl/common/cast.helper';
 
-interface CredentialOffer {
```
nishad-ayanworks, line 6: Could you please tell me why this is removed?

apps/api-gateway/src/issuance/dtos/issuance.dto.ts
```diff
@@ -118,18 +135,22 @@ export class OutOfBandCredentialDto {
     @IsEmail()
     @IsNotEmpty({ message: 'Please provide valid email' })
     @IsString({ message: 'email should be string' })
+    @Transform(({ value }) => value.trim())
```
nishad-ayanworks, line 138: Please also include ToLowercase() for the email.

nishad-ayanworks (commented):
Please refer and resolve the comments.

apps/api-gateway/src/issuance/dtos/issuance.dto.ts
bhavanakarwade: I have included toLowercase() for the email.

apps/api-gateway/src/issuance/dtos/issuance.dto.ts
bhavanakarwade: I removed the interface and created a DTO for it because I have to apply class validators to it.

PR #341, opened 2023-12-11
refactor: retrieve list of all schemas by network

bhavanakarwade:
## What ##
- Implement a function to retrieve a list of all schemas by network.

## Why ##
- To retrieve a list of all schemas by network is essential for seamless credential definition creation. 
- By having a comprehensive list of schemas available within the selected network, potential issues related to mismatched or unavailable schemas can be preemptively addressed. 
- This ensures that cred-def creation is accurate, preventing problems that may arise from incompatible schema information during the credential definition process.


KulkarniShashank (approved):
LGTM

PR #346, opened 2023-12-14
fix: pagination, sorting feature functionality and schema validations bug

bhavanakarwade:
### What ###

-  Resolved schema validation bugs during Schema Creation, including:
1.  Duplicate attribute names validation.
2. Case-insensitive duplicate attribute names validation.
3. Trim validations during attribute storage on the ledger and in the database.

- Refactor pagination and sorting features functionality.

### Why ###
- To ensure uniqueness, introduced case-insensitive validations for attribute names to prevent ambiguity, and incorporated trimming validations during attribute storage for consistent handling of leading and trailing spaces, enhancing data integrity.


nishad-ayanworks (approved):
LG

PR #364, opened 2023-12-20
refactor: Create Schema

nishad-ayanworks:
### WHAT ###
- Refactored the POST API for `/orgs/{orgId}/schemas`.
- Handled all the validations and negative scenarios.
- Refactored global exception handler.

### DIFF
--- a/libs/common/src/response-messages/index.ts
+++ b/libs/common/src/response-messages/index.ts
@@ -68,7 +68,8 @@ export const ResponseMessages = {
             userNotFound: 'User not found for the given organization',
             updateUserRoles: 'Unable to update user roles',
             deleteOrg: 'Organization not found',
-            notFound: 'Organization agent not found'
+            notFound: 'Organization agent not found',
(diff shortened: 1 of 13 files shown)


libs/common/src/response-messages/index.ts
```diff
@@ -68,7 +68,8 @@ export const ResponseMessages = {
             userNotFound: 'User not found for the given organization',
             updateUserRoles: 'Unable to update user roles',
             deleteOrg: 'Organization not found',
-            notFound: 'Organization agent not found'
+            notFound: 'Organization agent not found',
```
vivekayanworks, line 71: Replace "notFound" with more meaningful name.

vivekayanworks (approved):

libs/common/src/response-messages/index.ts
nishad-ayanworks: As discussed with @vivekayanworks. We will resolved this in subsequent PR

PR #370, opened 2023-12-21
refactor: webhook response for create connection API

bhavanakarwade:
### What ###
- Refactored the POST API ```wh/{orgId}/connections/```
2. Checked validations on payload.
3. Refactor response
4. Ensured consistency of API status codes.
5. Modified the NATS function.

### Why ###

### How ###
- Replaced the return type with a well-defined interface to enhance code clarity.
- Checked and standardized API status codes to ensure a uniform response.


apps/connection/src/connection.controller.ts
```diff
@@ -38,7 +38,7 @@ export class ConnectionController {
    * @returns Callback URL for connection and created connections details
    */
   @MessagePattern({ cmd: 'webhook-get-connection' })
-  async getConnectionWebhook(payload: IConnectionInterface): Promise<object> {
+  async getConnectionWebhook(payload: Connection): Promise<object> {
```
vivekayanworks, line 41: Replace object as a return type with appropriate interface.

apps/connection/src/connection.repository.ts
```diff
@@ -93,7 +93,7 @@ export class ConnectionRepository {
    * @returns Get connection details
    */
   // eslint-disable-next-line camelcase
-  async saveConnectionWebhook(payload: IConnectionInterface): Promise<object> {
+  async saveConnectionWebhook(payload: Connection): Promise<object> {
```
vivekayanworks, line 96: Replace object as a return type with appropriate interface.

apps/connection/src/connection.service.ts
```diff
@@ -113,7 +113,7 @@ export class ConnectionService {
    * @param user
    * @returns Connection legacy invitation URL
    */
-  async getConnectionWebhook(payload: IConnectionInterface): Promise<object> {
+  async getConnectionWebhook(payload: Connection): Promise<object> {
```
vivekayanworks, line 116: Replace object as a return type with appropriate interface.

apps/connection/src/connection.controller.ts
bhavanakarwade: It's not clear what data will come from the webhook, so its return type has been kept as an object. This was decided after a discussion with Ankita Ma'am and you, that for the webhook's APIs, the return type should remain as an object.

apps/connection/src/connection.service.ts
bhavanakarwade: It's not clear what data will come from the webhook, so its return type has been kept as an object. This was decided after a discussion with Ankita Ma'am and you, that for the webhook's APIs, the return type should remain as an object.

apps/connection/src/connection.repository.ts
bhavanakarwade: It's not clear what data will come from the webhook, so its return type has been kept as an object. This was decided after a discussion with Ankita Ma'am and you, that for the webhook's APIs, the return type should remain as an object.

apps/connection/src/interfaces/connection.interfaces.ts
```diff
@@ -51,12 +51,12 @@ export interface IOrgAgentInterface {
   orgId: string;
 }
 
-export class IConnectionInterface {
-  connectionDto: ConnectionPayload;
+export class ICreateConnection {
```
vivekayanworks, line 54: Please explain why are we using class instead of interface?

bhavanakarwade: I have changed class by interface.

vivekayanworks (approved):

PR #372, opened 2023-12-23
refactor: Connections by Organization

bhavanakarwade:
### What ###
- Refactored the GET API ```orgs/{orgId}/connections```
2. Ensured consistency of API status codes.
3. Modified the NATS function.
4. Tested with both positive and negative scenarios.

### Why ###

### How ###

-  Replaced the return type with a well-defined interface to enhance code clarity.
-  Checked and standardized API status codes to ensure a uniform response.

### DIFF
--- a/libs/common/src/interfaces/response.interface.ts
+++ b/libs/common/src/interfaces/response.interface.ts
@@ -5,3 +5,11 @@ export default interface IResponseType {
   data?: unknown;
   error?: unknown;
 };
+
+export default interface IResponse {
+  statusCode: number;
+  message?: string;
+  label?: string;
+  data?: unknown;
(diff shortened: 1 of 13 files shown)


libs/common/src/interfaces/response.interface.ts
```diff
@@ -5,3 +5,11 @@ export default interface IResponseType {
   data?: unknown;
   error?: unknown;
 };
+
+export default interface IResponse {
+  statusCode: number;
+  message?: string;
```
vivekayanworks, line 11: As discussed with team, let's remove error and make message mandatory and keep other fields as it is.

vivekayanworks (approved):

PR #382, opened 2023-12-27
refactor: create connection API

bhavanakarwade:
### What ###
- Refactored the POST API ```orgs/{orgId}/connections```
2. Checked validations on payload.
3. Refactor response.
4. Ensured consistency of API status codes.
5. Modified the NATS function.

### Why ###

### How ###
- Replaced the return type with a well-defined interface to enhance code clarity.
- Checked and standardized API status codes to ensure a uniform response.


apps/api-gateway/src/connection/dtos/connection.dto.ts
```diff
@@ -1,32 +1,32 @@
...
 
-    @ApiProperty()
+    @ApiPropertyOptional()
     @IsBoolean()
     @IsOptional()
     @IsNotEmpty({ message: 'please provide multiUseInvitation' })
     multiUseInvitation: boolean;
 
-    @ApiProperty()
+    @ApiPropertyOptional()
```
vivekayanworks, line 29: Replace "should" with "should be"

libs/common/src/interfaces/connection.interface.ts
```diff
@@ -19,4 +19,16 @@ export interface IConnectionsListCount {
     lastPage: number;
     data: IConnectionItem[];
   }
+
+  export interface ICreateConnectionUrl {
```
vivekayanworks, line 23: Why this interface is kept in libs and the other one is inside connection microservice?

bhavanakarwade: @vivekayanworks Sir, The interface is stored in the common libraries because I intend to use it in both the API Gateway and the Connection service. So as per our discussion, I've included it in the common libraries' interface file. The other interface, however, is exclusive to the Connection service, as I plan to use it only there.

apps/api-gateway/src/connection/dtos/connection.dto.ts
bhavanakarwade: Resolved.

vivekayanworks (approved):

PR #386, opened 2023-12-28
refactor: User module APIs.

pranalidhanavade:
- Refactored following APIs:
1. GET `/users/platform-settings` Get all platform and ecosystem settings.
2. PUT `/users/platform-settings` Update platform and ecosystem settings.
3. GET `/users/user-credentials` Get user credentials by Id.
4. POST `/users/certificate` Share user certificate

### What ###
1. Ensured consistency of API status codes.
2. Modified the NATS function.
5. Tested with both positive and negative scenarios.

### How ###
1. Checked and standardized API status codes to ensure a uniform response.
2. Modified the NATS function to accommodate the adjustments made in the API.

### Why ###
1. Ensuring consistent API status codes improves communication between different components of the system and enhances overall reliability.

### Extra points ###

- Implementation of well-defined interface by replacing previous existing Promise<object> is remaining for this APIs.


libs/common/src/response-messages/index.ts
```diff
@@ -18,7 +18,8 @@ export const ResponseMessages = {
             userCredentials: 'User credentials fetched successfully',
             platformEcosystemettings: 'Platform and ecosystem settings updated',
             fetchPlatformSettings: 'Platform settings fetched',
-            signUpUser:'User created successfully'
+            signUpUser:'User created successfully',
+            shareUserCertificate:'Certificate url generated successfully'
```
vivekayanworks, line 22: URL should be in capital letters.

apps/user/interfaces/user.interface.ts
```diff
@@ -113,10 +113,10 @@ export interface ISendVerificationEmail {
   }
   
   export interface ICheckUserDetails {
```
vivekayanworks, line 115: Can we reduce no. of fields as discussed earlier?

libs/common/src/response-messages/index.ts
pranalidhanavade: Resolved.

apps/user/interfaces/user.interface.ts
pranalidhanavade: Yes sir.

pranalidhanavade: Resolved.

vivekayanworks (approved):

PR #393, opened 2023-12-29
fix: create service name random in shell script

KulkarniShashank:
### What ###
Create a random service name within the shell script. It eliminates unnecessary parameters related to agent provisioning.

### Why ###
The current script includes unceremonious parameters and agent-provisioning, which are redundant and can lead to issues. This fix aims to streamline the script by removing these unnecessary elements.

### How ###
Script to generate a random service name without the need for explicit parameters and eliminating the agent-provisioning code. The updated script is designed to improve simplicity and efficiency.


Sheetal-ayanworks (approved):
LG

PR #398, opened 2023-12-29
feat: Set organization to public by default 

nishad-ayanworks:
### What ###
- Set organization to public by default


bhavanakarwade (approved):
LG

PR #415, opened 2024-01-04
fix: solved the bug for sign and submit endorsement 

KulkarniShashank:
### What ###

1. Resolved the issue related to the "apiKey not found" error when attempting to sign and submit an endorsement.

### Why ###

1. The problem was occurring due to the absence of the apiKey during the process of signing and submitting an endorsement. This was preventing successful completion of the operation.

### How ###

1. Identified the root cause of the issue, which was the missing apiKey during the sign and submit endorsement process.
2. Implemented a fix to ensure that the apiKey is properly included and accessible during the operation.
3. Tested the solution to confirm that the "apiKey not found" issue is resolved.


nishad-ayanworks (approved):
LG

PR #416, opened 2024-01-04
refactor: get user details API

pranalidhanavade:
### What 

- Update response for an unregistered, verified email while fetching user details.
- Populate only necessary fields in the API response.
- Handled error responses and status codes.

### DIFF
--- a/apps/api-gateway/src/dtos/email-validator.dto.ts
+++ b/apps/api-gateway/src/dtos/email-validator.dto.ts
@@ -1,4 +1,5 @@
 
+import { trim } from '@credebl/common/cast.helper';
 import { ApiProperty } from '@nestjs/swagger';
 import { Transform } from 'class-transformer';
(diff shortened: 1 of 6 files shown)


ankita-p17 (approved):

apps/api-gateway/src/dtos/email-validator.dto.ts
```diff
@@ -9,5 +10,6 @@ export class EmailValidator {
   @IsNotEmpty({ message: 'Email is required' })  
   @MaxLength(256, { message: 'Email must be at most 256 character' })
   @IsEmail()
+  @Transform(({ value }) => 'string' === typeof value ? value.trim() : value)
```
nishad-ayanworks, line 13: Please use the `trim()` function of `cast.helper file`

nishad-ayanworks (changes requested):
Please resolve comment

PR #417, opened 2024-01-04
fix: solved the agent spin-up with apiKey updation on org_agent table

KulkarniShashank:
### What ###
Resolves the agent spin-up issue by implementing a crucial apiKey update in the org_agent table.

### Why ###
The agent spin-up problem was identified to be linked to outdated API keys stored in the org_agent table. This update aims to ensure that the API keys are consistently updated, thus eliminating the spin-up issue.

### How ###
The update involves making necessary modifications to the org_agent table, ensuring that the apiKey field is updated appropriately during the agent spin-up process. This will result in improved functionality and a more reliable agent deployment process.


ankita-p17:
Lgtm

PR #420, opened 2024-01-05
fix: Invalid invitation status transition

KulkarniShashank:
### what ###

- Resolved the status update issue in organization service

### why ###

- User was unable to accept/reject the Invitation to join an organization

### How ###

- Added transition function and array for valid status updates and applied condition to check status update in organization service using the same


ankita-p17 (approved):
LGTM

PR #435, opened 2024-01-13
feat: Feat/oob credential issuance verification

ankita-p17:
## What?
- New API to create out-of-band credential offer
- Remove static status codes from all the files from the repository

## Why?
- Organization should be able to create out-of-band credential offer

## How?
- New API in api-gateway
- New endpoint in issuance service


KulkarniShashank (approved):
LG

PR #476, opened 2024-02-01
Feat/oob credential issuance verification

ankita-p17:
### What ###
- Added schema as restriction in proof request
- Refactored oob proof request create API to make email optional

### Why ###
 - OOB proof request API should be generic. If email is provided it should send the request on email other API should return proof-request payload

### How ###
- Added validator on email attribute
- Added schema_id as a restriction


amitpadmani-awts (approved):
LGFM

PR #489, opened 2024-02-06
fix: Passkey fixes

nishad-ayanworks:
### What ###

- Worked on the passkey fixes for keycloak reset password registration.

### DIFF
--- a/apps/user/src/user.service.ts
+++ b/apps/user/src/user.service.ts
@@ -296,6 +296,13 @@ export class UserService {
         throw new NotFoundException(ResponseMessages.user.error.emailNotVerified);
       }
 
+      const decryptedPassword = await this.commonService.decryptPassword(userInfo.password);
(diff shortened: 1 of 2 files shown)


apps/user/src/user.service.ts
```diff
@@ -296,6 +296,13 @@ export class UserService {
         throw new NotFoundException(ResponseMessages.user.error.emailNotVerified);
       }
 
+      const decryptedPassword = await this.commonService.decryptPassword(userInfo.password);
```
tipusinghaw, line 299: should we generate the hash of the password and send it to the keycloak?

KulkarniShashank (approved):
LG

apps/user/src/user.service.ts
nishad-ayanworks: Right now we are directly passing the password to the keycloak as it is.

PR #493, opened 2024-02-07
feat: forgot password

nishad-ayanworks:
### What ###

- Created a migration for `reset password token`
- Worked on the forgot and reset password functionality.


KulkarniShashank (approved):
LG

PR #494, opened 2024-02-08
fix: platform not allowing endorsement request with the same schema and version

pranalidhanavade:
### What 

- Fixed issue platform not allowing endorsement request with the same schema and version


nishad-ayanworks (approved):
LG

PR #495, opened 2024-02-08
fix: removed cred def id and schema id restrictions

bhavanakarwade:
### What ### 
- Removed cred def and schema restrictions while sending proof requests.


nishad-ayanworks (approved):
LG

KulkarniShashank (approved):
LG

PR #498, opened 2024-02-09
fix: user registration for keycloak

nishad-ayanworks:
### What ###

-  fixes of user registration for keycloak


pranalidhanavade (approved):
LG

PR #500, opened 2024-02-09
fix: unauthorized error message

sanjay-k1910:
## What?
Updated unauthorized error message when a user tries to use API with the invalid access token
Previous response:
```
{
  "statusCode": 500,
  "message": "Cannot read properties of null (reading 'iss')",
  "error": "Internal Server error"
}
```
Refactored response:
```
{
  "statusCode": 401,
  "message": "Invalid access token",
  "error": "Unauthorized"
}
```

## Why?
If an unauthorized user tries to access API then it should return a proper error message.

### DIFF
--- a/libs/common/src/response-messages/index.ts
+++ b/libs/common/src/response-messages/index.ts
@@ -59,8 +59,8 @@ export const ResponseMessages = {
             invitationStatusUpdateInvalid: 'Status update is invalid. Request is already',
             resetSamePassword: 'New password should not be the current password',
             resetPasswordLink: 'Unable to create reset password token',
-            invalidResetLink: 'Invalid reset password link or expired',
-            invalidAccessToken: 'Invalid access token'
+            invalidResetLink: 'Invalid or expired reset password link',
+            invalidAccessToken: 'Authentication failed'
         }
     },
     organisation: {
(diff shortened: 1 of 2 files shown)


libs/common/src/response-messages/index.ts
```diff
@@ -59,7 +59,8 @@ export const ResponseMessages = {
             invitationStatusUpdateInvalid: 'Status update is invalid. Request is already',
             resetSamePassword: 'New password should not be the current password',
             resetPasswordLink: 'Unable to create reset password token',
-            invalidResetLink: 'Invalid reset password link or expired'
+            invalidResetLink: 'Invalid reset password link or expired',
```
ankita-p17, line 62: update verbose
"Invalid or expired reset password link"

libs/common/src/response-messages/index.ts
```diff
@@ -59,7 +59,8 @@ export const ResponseMessages = {
...
+            invalidAccessToken: 'Invalid access token'
```
ankita-p17, line 63: 'Unauthorized' or 'Authentication failed' message is also fine. No need to provide technical details to user.

libs/common/src/response-messages/index.ts
sanjay-k1910: Resolved.

sanjay-k1910: Resolved.

libs/common/src/response-messages/index.ts
sanjay-k1910: Resolved.

ankita-p17 (approved):

PR #519, opened 2024-02-16
feat: Included the format data within the proof's webhook

pallavighule:
### What ###

1. Included the format data within the proof's webhook


KulkarniShashank (approved):
LG

PR #520, opened 2024-02-19
feat: Create new connection invitation

bhavanakarwade:
### What
- Modify the create invitation function, So that it checks if the Organization logo and Organization name have been updated, then a new connection invitation will be created.

### Why
- This modification ensures that invitations accurately reflect the current changes of the organization. If the organization's logo or name changes, it's essential that invitations reflect these updates.

### DIFF
--- a/apps/organization/src/organization.service.ts
+++ b/apps/organization/src/organization.service.ts
@@ -294,7 +294,7 @@ export class OrganizationService {
     logoUrl: string,
     orgId: string
   ): Promise<ICreateConnectionUrl> {
-    const pattern = { cmd: 'create-connection1' };
+    const pattern = { cmd: 'create-connection' };
 
     const payload = {
       orgName,
--- a/apps/organization/src/organization.service.ts
+++ b/apps/organization/src/organization.service.ts
@@ -248,13 +248,10 @@ export class OrganizationService {
   async updateOrganization(updateOrgDto: IUpdateOrganization, userId: string, orgId: string): Promise<organisation> {
     try {
 
-      const organizationExist = await this.organizationRepository.checkOrganizationExist(updateOrgDto.name, orgId);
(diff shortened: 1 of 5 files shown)


apps/organization/src/organization.service.ts
```diff
@@ -250,7 +250,7 @@ export class OrganizationService {
 
       const organizationExist = await this.organizationRepository.checkOrganizationExist(updateOrgDto.name, orgId);
```
ankita-p17, line 251: you can optimize this code
1. get org by name
2. check if org.Id received is same as orgId, if true, throw new ConflictException(ResponseMessages.organisation.error.exists);

ankita-p17 (changes requested):

apps/organization/src/organization.service.ts
```diff
@@ -268,7 +268,18 @@ export class OrganizationService {
         delete updateOrgDto.logo;
       }
 
-      const organizationDetails = await this.organizationRepository.updateOrganization(updateOrgDto);
+      let organizationDetails;
+      const checkAgentIsExists = await this.organizationRepository.getAgentInvitationDetails(orgId);
+
+      if (!checkAgentIsExists?.connectionInvitation && !checkAgentIsExists?.agentId) {
```
ankita-p17, line 274: Below these two if conditions to optimize the code and avoid checking same condition repeatedly

ankita-p17 (changes requested):

apps/organization/src/organization.service.ts
bhavanakarwade: Resolved

apps/organization/src/organization.service.ts
bhavanakarwade: Resolved

ankita-p17 (approved):

PR #528, opened 2024-02-20
fix: issuance API with required attribute

pranalidhanavade:
### What 

- Fixed issue regarding issuance API with old schema details.

### DIFF
--- a/apps/issuance/src/issuance.service.ts
+++ b/apps/issuance/src/issuance.service.ts
@@ -144,25 +144,24 @@ export class IssuanceService {
 
       const hasRequiredAttributes = attributes?.some(attribute => attribute.isRequired);
 
-      if (hasRequiredAttributes) {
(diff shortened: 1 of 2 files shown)


apps/issuance/src/issuance.service.ts
```diff
@@ -133,28 +144,37 @@ export class IssuanceService {
...
+        attributesArray.forEach((attribute) => {
+          if (attribute.attributeName && attribute.isRequired) {
             
-            return attribute.isRequired && !attribute.value;
-          });
-      
-        if (0 < attrError.length) {
-          throw new BadRequestException(attrError);
-        }
+            payload.attributes.map((attr) => {
```
ankita-p17, line 162: @pranalidhanavade It would be better to create a common function for this logic and call from wherever required

ankita-p17 (approved):
The added comment will be resolved in next PR soon

PR #529, opened 2024-02-21
fix: user friendly message of ecosystem roles if user does not have access

sanjay-k1910:
## What?
- Updated error message if the user does not have access to call an API.
- Updated message from "Forbidden resource" to "User does not have access".

## Why?
To provide user-friendly message


ankita-p17 (approved):

nishad-ayanworks (approved):
LG

PR #532, opened 2024-02-22
fix: ecosystem member list search sorting issue

sanjay-k1910:
## What?
- Resolved the issue of searching and sorting
- Updated query to sort value as per the column name

## Why?
- On the studio side, search and sort should work as expected to improve user accessibility for the records

### DIFF
--- a/apps/api-gateway/src/authz/guards/ecosystem-roles.guard.ts
+++ b/apps/api-gateway/src/authz/guards/ecosystem-roles.guard.ts
@@ -50,17 +50,17 @@ export class EcosystemRolesGuard implements CanActivate {
       const ecosystemOrgData = await this.ecosystemService.fetchEcosystemOrg(ecosystemId, orgId);
 
       if (!ecosystemOrgData) {
-        throw new ForbiddenException('Organization does not match');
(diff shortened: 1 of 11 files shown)


apps/api-gateway/src/authz/guards/ecosystem-roles.guard.ts
```diff
@@ -54,19 +50,17 @@ export class EcosystemRolesGuard implements CanActivate {
       const ecosystemOrgData = await this.ecosystemService.fetchEcosystemOrg(ecosystemId, orgId);
 
       if (!ecosystemOrgData) {
-        throw new HttpException('Organization does not match', HttpStatus.FORBIDDEN);
+        throw new ForbiddenException('Organization does not match');
```
nishad-ayanworks, line 53: Please add these messages in the common file

nishad-ayanworks (approved):
LG

apps/api-gateway/src/authz/guards/ecosystem-roles.guard.ts
sanjay-k1910: Resolved.

PR #535, opened 2024-02-22
fix: user registration restrictions

nishad-ayanworks:
### What ###

- Added restriction of user registration if already registered on `keycloak` either on `supabase`


sanjay-k1910 (approved):
LGTM!

PR #538, opened 2024-02-23
Remove unwanted attributes, Added support for more restrictions

ankita-p17:
- Removed attributes which are not needed the API
- Added descriptive example


amitpadmani-awts (approved):
LGTM

vivekayanworks (approved):

PR #545, opened 2024-02-26
fix: extra attribute in issuance payload

pranalidhanavade:
### What 

- Removed extra attribute from issuance payload.


KulkarniShashank (approved):
LG

PR #547, opened 2024-02-27
fix: solved verification predicates values

KulkarniShashank:
### What ###

Resolved verification predicates values to ensure accurate validation.
### Why ###

The verification predicates were causing inconsistencies, and this fix ensures the correctness of validation processes.
### How ###

Updated the relevant code to address and resolve the verification predicates values.
Made changes to ensure accurate validation during the verification process.


nishad-ayanworks (approved):
LG

PR #552, opened 2024-02-28
fix: solved issue for the ledgerId restriction as a string of array

KulkarniShashank:
### What ###
Resolved the issue related to the ledgerId restriction when passed as a string of an array.

### Why ###
The existing implementation did not handle the ledgerId restriction properly when provided as a string within an array. This caused unexpected behavior and potential errors in certain scenarios. The fix ensures that the ledgerId restriction is correctly processed when presented in this format.

### How ###
Modified the relevant function to correctly parse and handle ledgerId when passed as a string within an array.
Added test cases to cover the scenarios where the issue was identified.
Updated documentation to reflect the corrected behavior and provide examples of using ledgerId in this specific format.

### DIFF
--- a/apps/api-gateway/src/agent-service/dto/create-tenant.dto.ts
+++ b/apps/api-gateway/src/agent-service/dto/create-tenant.dto.ts
@@ -25,11 +25,13 @@ export class CreateTenantDto {
     })
     seed: string;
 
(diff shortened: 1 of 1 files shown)


nishad-ayanworks (approved):
LG

PR #553, opened 2024-02-28
refactor: common service & client service

nishad-ayanworks:
### What ###

- Refactored the common service and logs
- Refactored the client service and logs


KulkarniShashank (approved):
LG

PR #557, opened 2024-02-29
fix: error handling for endorsement flow

pranalidhanavade:
### What 

- Fixed issue:- Incorrect message displayed when submitting schema with the same name and version that is already submitted on ledger (Endorsement flow).
-  Removed unnecessary logs from issuance and verification services.


KulkarniShashank (approved):
LG

PR #565, opened 2024-03-04
fix: solved bulk issuance and with email issuance flow 

KulkarniShashank:
- Solved bulk issuance when upload csv and issue the credentials

### DIFF
--- a/apps/issuance/src/issuance.service.ts
+++ b/apps/issuance/src/issuance.service.ts
@@ -9,7 +9,7 @@ import { ResponseMessages } from '@credebl/common/response-messages';
 import { ClientProxy, RpcException } from '@nestjs/microservices';
 import { map } from 'rxjs';
 // import { ClientDetails, FileUploadData, ICredentialAttributesInterface, ImportFileDetails, OutOfBandCredentialOfferPayload, PreviewRequest, SchemaDetails } from '../interfaces/issuance.interfaces';
-import { FileUploadData, IAttributes, IClientDetails, ICreateOfferResponse, IIssuance, IIssueData, IPattern, ISendOfferNatsPayload, ImportFileDetails, IssueCredentialWebhookPayload, OutOfBandCredentialOfferPayload, PreviewRequest, SchemaDetails } from '../interfaces/issuance.interfaces';
+import { CredentialOffer, FileUploadData, IAttributes, IClientDetails, ICreateOfferResponse, IIssuance, IIssueData, IPattern, ISendOfferNatsPayload, ImportFileDetails, IssueCredentialWebhookPayload, OutOfBandCredentialOfferPayload, PreviewRequest, SchemaDetails, SendEmailCredentialOffer } from '../interfaces/issuance.interfaces';
 import { OrgAgentType } from '@credebl/enum/enum';
 // import { platform_config } from '@prisma/client';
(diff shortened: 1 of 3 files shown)


shitrerohit (approved):

ankita-p17 (approved):

apps/issuance/src/issuance.service.ts
```diff
@@ -425,259 +426,305 @@ export class IssuanceService {
...
+  const iterationNo = index + 1;
+  try {
+    let outOfBandIssuancePayload;
+    if (IssueCredentialType.INDY === credentialType) {
+    
+      outOfBandIssuancePayload = {
+        protocolVersion: protocolVersion || 'v1',
+        credentialFormats: {
+          indy: {
+            // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
tipusinghaw, line 630: any specific reason to disable eslint

tipusinghaw, line 630: Any specific reason to disable eslint?

apps/issuance/src/issuance.service.ts
```diff
@@ -425,259 +426,305 @@ export class IssuanceService {
...
+          sendEmailCredentialOffer['emailId'] = iterator.emailId;
+          sendEmailCredentialOffer['index'] = index;
 
-          if (!invitationId) {
-            errors.push(new NotFoundException(ResponseMessages.issuance.error.invitationNotFound));
-            return false;
-          }
+          return this.sendEmailForCredentialOffer(sendEmailCredentialOffer);
+        });
+        emailPromises.push(Promise.all(batchPromises));
```
tipusinghaw, line 560: I am not sure how it will handle the remaining batch if the 1st batch will fail.

apps/issuance/src/issuance.service.ts
```diff
@@ -425,259 +426,305 @@ export class IssuanceService {
...
+      }
+    } else {
+      emailPromises.push(this.sendEmailForCredentialOffer(sendEmailCredentialOffer));
+    }
 
-          const agentEndPoint = agentDetails.tenantId
-            ? `${agentDetails.agentEndPoint}/multi-tenancy/url/${agentDetails.tenantId}/${invitationId}`
-            : `${agentDetails.agentEndPoint}/url/${invitationId}`;
+    const results = await Promise.all(emailPromises);
```
tipusinghaw, line 566: After storing all the values, again we are passing the array in promise.

apps/issuance/src/issuance.service.ts
KulkarniShashank: We need to eliminate the isRequired attribute from the payload to avoid mapping unnecessary attributes as required. Therefore, we should create a separate variable for isRequired and ensure that it is not utilized within this function.

apps/issuance/src/issuance.service.ts
KulkarniShashank: Resolved

apps/issuance/src/issuance.service.ts
KulkarniShashank: We will check the possibilities for validation and we will create separate PR for it.

shitrerohit (approved):

shitrerohit (approved):

PR #568, opened 2024-03-05
fix: org client validations & org create validations

nishad-ayanworks:
### What ###

- Fixed validation on org create when org slug matches.
- Refactor the user and org client login validations.


tipusinghaw (approved):
LGTM

shitrerohit (approved):

PR #581, opened 2024-03-07
fix: solved the endosement credential definition

KulkarniShashank:
### What ###
Resolved the endorsement credential definition.
Added the keycloak configuration in .env.sample file.

### Why ###
When submit the credential definition getting error schema name not found.

### How ###
Added the validation if getting the type as a schema on the endorsement.


nishad-ayanworks (approved):
LG

shitrerohit (approved):

PR #583, opened 2024-03-08
fix: remove unnecessary logs

GHkrishna:
**What**?
Remove Unnecessary logs from being logged.

**Where?**
In Issuance service


KulkarniShashank (approved):
LGTM

PR #589, opened 2024-03-11
feat: support multiple did methods

bhavanakarwade:
### What?
- Implementation of multiple DID methods support during wallet creation
- Segregation of create wallet and create DID functionality.

### Why?
- A single endpoint ensures the generality of DID creation.


apps/api-gateway/src/agent-service/agent-service.controller.ts
```diff
@@ -172,4 +190,95 @@ export class AgentController {
...
+   * @param orgId 
+   * @returns wallet
+   */
+     @Post('/orgs/:orgId/agents/createWallet')
+     @ApiOperation({
+       summary: 'Create wallet',
+       description: 'Create wallet'
+     })
+     @UseGuards(AuthGuard('jwt'), OrgRolesGuard)
+     @Roles(OrgRoles.OWNER, OrgRoles.ADMIN)
```
tipusinghaw, line 205: Add role for `issuer` as well

tipusinghaw (approved):

PR #592, opened 2024-03-13
fix: refactored tables

bhavanakarwade:
### What?
- Refactored JSON data for both the ledgerConfig and ledgers table


tipusinghaw (approved):
LG

PR #593, opened 2024-03-13
feat: Add reuse connection

pallavighule:
### What ###

- Add reuse connection functionality


ankita-p17 (approved):
Approved with verbose changes communicated

PR #594, opened 2024-03-13
feat: Auth refresh token

nishad-ayanworks:
### What ###

- Implemented POST API of `/auth/refresh-token` to get the `access_token` from the `refresh_token`

### Why ###

- To get the updated  `access_token` if current is `expired`


KulkarniShashank (approved):
LG

tipusinghaw (approved):

PR #595, opened 2024-03-13
fix: delete the ecosystem invitation which is already deleted

pranalidhanavade:
### What 

- Fixed issue: An incorrect message appears when delete the ecosystem invitation which is already deleted


KulkarniShashank (approved):
LG

PR #656, opened 2024-04-12
fix: API error for get credef by id

GHkrishna:
Fixes get API Key not found error while getting cred-def by id 

## What:
Fixes API Key error while getting credential definition by id


tipusinghaw (approved):
LGTM

PR #665, opened 2024-04-15
hotfix: event certificate share option

tipusinghaw:
#### What ####
- Added option to share event certificates
#### How ####
- Added `template` for pinnacle and other certificates
- Added Qr-code generation logic

### DIFF
--- a/libs/aws/src/aws.service.ts
+++ b/libs/aws/src/aws.service.ts
@@ -42,14 +42,14 @@ export class AwsService {
 
     try {
       await putObjectAsync({
-        Bucket: `${process.env.AWS_ORG_LOGO_BUCKET_NAME}`,
-        Key: `${pathAWS}/${encodeURIComponent(filename)}-${timestamp}.png`,
+        Bucket: `${bucketName}`,
+        Key: `${pathAWS}/${encodeURIComponent(filename)}-${timestamp}.${ext}`,
         Body: fileBuffer,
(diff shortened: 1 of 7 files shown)


libs/aws/src/aws.service.ts
```diff
@@ -42,14 +42,14 @@ export class AwsService {
...
-        Key: `${pathAWS}/${encodeURIComponent(filename)}-${timestamp}.png`,
+        Bucket: `${bucketName}`,
+        Key: `${pathAWS}/${encodeURIComponent(filename)}-${timestamp}.${ext}`,
         Body: fileBuffer,
         ContentEncoding: encoding,
         ContentType: `image/png`
       });
 
-      const imageUrl = `https://${process.env.AWS_ORG_LOGO_BUCKET_NAME}.s3.${process.env.AWS_PUBLIC_REGION}.amazonaws.com/${pathAWS}/${encodeURIComponent(filename)}-${timestamp}.${ext}`;
+      const imageUrl = `https://${bucketName}.s3.${process.env.AWS_PUBLIC_REGION}.amazonaws.com/${pathAWS}/${encodeURIComponent(filename)}-${timestamp}.${ext}`;
```
GHkrishna, line 52: Might cause issue if the bucketName contains dot(.)

However as discussed with @tipusinghaw this is not the case for us.

GHkrishna (approved):

PR #678, opened 2024-04-24
Update presentation definition payload of W3C request

ankita-p17:
##What
Added a missing attribute `name` in presentationDefinition payload. This is a shared attributes between the verifier and prover addition to the requested data attributes.

##Why
To be aligned with the standard specification.


amitpadmani-awts (approved):
LGTM

PR #689, opened 2024-04-29
refactor: remove legacy connection invitation

KulkarniShashank:
### What ###
Remove the legacy connection invitation feature from the codebase.

### Why ###
The legacy connection invitation feature is no longer necessary.

### How ###
Removes all code related to the legacy connection invitation feature


pallavighule (approved):
LGTM

PR #690, opened 2024-04-30
Feat/verification w3c with connection

GHkrishna:
## What
Change in request payload for verification with connection.

## Why
Adds support for w3 verification with connection


pallavighule (approved):
LGTM

PR #691, opened 2024-04-30
fix: set primary did issue

tipusinghaw:
#### What ####
- stored the did while spinning up the agnet
- fixed the API method issue


bhavanakarwade (approved):
LG

KulkarniShashank (approved):
LG

PR #693, opened 2024-04-30
feat: w3c credential issuance by connection id

bhavanakarwade:
### What?
- Worked on w3c credential issuance by connection Id (support multiple connection id selection)

### Why?
- To support w3c credential type while issuance


pallavighule (approved):
LGTM

PR #706, opened 2024-05-04
fix: store did document while creating did

tipusinghaw:
#### What ####
- Fixed store did document issue while creating did
- Removed data from set primary did API
- Added did document in `org_agent` while changing primary did


bhavanakarwade (approved):
LG

PR #713, opened 2024-05-13
feat: on-premise agent configure

KulkarniShashank:
### What ###
Implementing the Agent configure API, which allows for the configuration of agents within the system.

### Why ###
The Agent configure API is essential for enabling dynamic configuration of agents, providing flexibility and adaptability to changing requirements.

### How ###
Define the endpoints and payloads required for the Agent configure API.
Implement backend logic to process configuration requests and update agent settings accordingly.


libs/common/src/cast.helper.ts
```diff
@@ -130,16 +131,32 @@ export class ImageBase64Validator implements ValidatorConstraintInterface {
...
\ No newline at end of file
+
+export class AgentSpinupValidator {
+  private static validateField(value: string, errorMessage: string): void {
+      if (!value) {
+          throw new BadRequestException(errorMessage);
+      }
+  }
+
+  private static validateWalletName(walletName: string): void {
```
bhavanakarwade, line 142: There is no need to validate the wallet name, as the agent accepts alphanumeric characters

bhavanakarwade (changes requested):

bhavanakarwade (approved):
LG

tipusinghaw (approved):

PR #731, opened 2024-05-22
refactor: on premises agent configure

KulkarniShashank:
### What ###
Refactored the configuration process for the on-premises agent.

### Why ###
Improved and added the .educreds folder for the store configuration

### How ###

- Reorganized configuration files for better clarity.
- Implemented new configuration parameters to enhance flexibility.
- Optimized existing code to reduce redundancy.


apps/agent-service/src/agent-service.service.ts
```diff
@@ -282,15 +281,14 @@ export class AgentServiceService {
         orgId,
         agentEndPoint: agentEndpoint,
         orgAgentTypeId,
-        ledgerId: ledgerIdData.map((ledger) => ledger?.id),
+        ledgerId: null,
```
tipusinghaw, line 284: any specific reason to assign null?

apps/agent-service/src/interface/agent-service.interface.ts
```diff
@@ -37,7 +37,6 @@ export interface IAgentConfigure {
   apiKey: string;
   orgId: string;
   network?: string;
-  tenant?: boolean;
```
tipusinghaw, line 40: Please make sure this will not break tenancy flow

apps/agent-service/src/repositories/agent-service.repository.ts
```diff
@@ -418,4 +435,33 @@ export class AgentServiceRepository {
       throw error;
     }
   }
+
+  async getLedgerByNameSpace(indyNamespace: string): Promise<{
```
tipusinghaw, line 439: Please create a separate interface for return type

apps/agent-service/src/repositories/agent-service.repository.ts
```diff
@@ -418,4 +435,33 @@ export class AgentServiceRepository {
...
+    poolConfig: string;
+    isActive: boolean;
+    networkString: string;
+    nymTxnEndpoint: string;
+    indyNamespace: string;
+    networkUrl: string;
+}> {
+    try {
+      if (indyNamespace) {
+        const ledgerDetails = await this.prisma.ledgers.findFirst({
```
tipusinghaw, line 454: Please avoid using `findFirst`. I will return the 1st value if the namespace does not match

apps/agent-service/src/interface/agent-service.interface.ts
KulkarniShashank: Yes I tested and tenantId by default taking null value

tipusinghaw (approved):

PR #737, opened 2024-05-25
fix: Invalid nested response

GHkrishna:
## What:
- Fixes invalid response inside response for oob issuance

## Why:
- Previous response included response inside a response object

Fixes: [#406](https://github.com/credebl/automation-test-suite/issues/406)


bhavanakarwade (approved):
LG

PR #738, opened 2024-05-28
feat: bulk issuance implementation using nest queue

tipusinghaw:
### What ###
- Added nest queue `addBulk` method to add bulk payload
- Changed the endpoint for fetching all template details and downloading CSV template
- Added W3C support for all the APIs


apps/api-gateway/src/issuance/issuance.controller.ts
```diff
@@ -172,28 +174,60 @@ export class IssuanceController {
...
+    @Res() res: Response,
+    @Query('schemaType') schemaType: SchemaType = SchemaType.INDY
+  ): Promise<Response> {
+    const templateList = await this.issueCredentialService.getAllCredentialTemplates(orgId, schemaType);
+    const credDefResponse: IResponse = {
+      statusCode: HttpStatus.OK,
+      message: ResponseMessages.credentialDefinition.success.template,
+      data: templateList
+    };
+    return res.status(HttpStatus.CREATED).json(credDefResponse);
```
bhavanakarwade, line 201: The res.status should be HttpStatus.OK for the GET API

apps/api-gateway/src/issuance/issuance.service.ts
```diff
@@ -114,7 +119,7 @@ export class IssuanceService extends BaseService {
         return this.sendNats(this.issuanceProxy, 'issued-file-data', payload);
     }
 
-    async issueBulkCredential(requestId: string, orgId: string, clientDetails: ClientDetails, reqPayload: RequestPayload): Promise<object> {
```
bhavanakarwade, line 117: Any reason to remove type `RequestPayload` ?

apps/issuance/src/issuance.service.ts
```diff
@@ -808,16 +803,45 @@ async sendEmailForCredentialOffer(sendEmailCredentialOffer: SendEmailCredentialO
...
+         fileName = `${schemaResponse.tag}-${timestamp}.csv`;
+
+      } else if (schemaType === SchemaType.W3C_Schema) {
+        const schemDetails = await this.issuanceRepository.getSchemaDetailsBySchemaIdentifier(templateId);
+        const {attributes, schemaLedgerId, name} = schemDetails;
+        schemaResponse.attributes = attributes;
+        schemaResponse.schemaLedgerId = schemaLedgerId;
+        schemaResponse.name = name;
+
+        if (!schemaResponse) {
```
bhavanakarwade, line 834: Just a thought, can we use this validation for both? We've already considered schemaResponse as a let, so can we use this block of code outside this condition?

libs/common/src/interfaces/issuance.interface.ts
```diff
@@ -15,4 +15,15 @@ export interface IIssuedCredential {
     previousPage: number;
     lastPage: number;
     data: IIssuedCredentialResponse[];
+  }
+  interface CredentialData {
```
bhavanakarwade, line 19: We are using the I prefix for interface names

apps/issuance/src/issuance.service.ts
tipusinghaw: After the validation, I assigned file name and the file name will be different for both the schema type

libs/common/src/interfaces/issuance.interface.ts
tipusinghaw: I am not exporting the `CredentialData` it's for internal use

bhavanakarwade (approved):

KulkarniShashank (approved):
LG

PR #746, opened 2024-06-03
feat : Added support for different email service provider.

ganeshawle25:
### What ###

- Added support for different email provider like Amazon SES, SMTP(for gmail) and refactored existing SendGrid functionality.

### Why ###

- As per client's requirement provide email support for Amazon SES and SMTP providers.

### How ###

- Added 'email.service.ts' class, which contains common 'sendEmail' as entry function and according to EMAIL_PROVIDER property it will call respective functionalities.
- Added new 'sendWithSES' and 'sendWithSMTP functions for each email provider.
- Refactored existing sendWithSendGrid function.


libs/common/src/aws-ses-helper-file.ts
```diff
@@ -0,0 +1,37 @@
...
+import { EmailDto } from './dtos/email.dto';
+import { Logger } from '@nestjs/common';
+import * as nodemailer from 'nodemailer';
+import * as AWS from 'aws-sdk';
+
+dotenv.config();
+
+export const sendWithSES = async (emailDto: EmailDto): Promise<boolean> => {
+  const awsSES = new AWS.SES({
+    apiVersion: '2010-12-01',
```
tipusinghaw, line 11: Is it possible to get this value from `.env` as well?

libs/common/src/aws-ses-helper-file.ts
```diff
@@ -0,0 +1,37 @@
...
+    }
+  });
+  const transporter = nodemailer.createTransport({
+    SES: awsSES
+  });
+  try {
+    return await transporter
+      .sendMail({
+        from: emailDto.emailFrom,
+        to: emailDto.emailTo, //list  of receivers
```
tipusinghaw, line 25: Please add comments where it isn't easy to under the flow

libs/common/src/email.service.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  constructor() {}
+  async sendEmail(emailDto: EmailDto): Promise<boolean> {
+    const provider = process.env.EMAIL_PROVIDER;
+    let result: boolean;
+    if (!provider) {
+      throw new Error('Email provider is not set in environment variables. Please provide email service provider property.');
+    }
+    try {
+      this.logger.debug(`Email Provider is ::::::::${provider}`);
```
tipusinghaw, line 19: Please remove this logger if it's not required.

tipusinghaw (approved):

PR #747, opened 2024-06-03
feat: bulk issuance retry

tipusinghaw:
#### What ####
- Implemented the bulk-issuance retry
- Added logic for last element of the array


KulkarniShashank (approved):
LG

libs/common/src/response-messages/index.ts
```diff
@@ -414,7 +414,8 @@ export const ResponseMessages = {
     },
     bulkIssuance: {
         success: {
-            create: 'Issuance process successfully'
+            create: 'Issuance process successfully',
```
bhavanakarwade, line 417: Make this response message informative

tipusinghaw: sure, thanks

bhavanakarwade (approved):
LGTM

PR #754, opened 2024-06-06
fix: update DID Doc by dedicated agent

KulkarniShashank:
### What ###

- Update the org_agent table to reflect the DID Doc when a DID is set as a primary by a dedicated agent.

### Why ###

- To maintain accurate and up-to-date information in the org_agent table.
- To ensure that the primary DID status is correctly represented, which is crucial for organizational operations and integrations.

### How ###

- Modify the logic in the agent's code to trigger an update in the org_agent table when a DID is set as a primary.
- Add a database query to update the DID Doc in the org_agent table based on the dedicated agent's action.

### DIFF
--- a/apps/agent-service/src/agent-service.service.ts
+++ b/apps/agent-service/src/agent-service.service.ts
@@ -991,8 +991,8 @@ export class AgentServiceService {
           description: ResponseMessages.errorMessages.serverError
         });
       }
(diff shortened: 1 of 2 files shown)


bhavanakarwade (approved):
LG

bhavanakarwade (approved):
LG

PR #756, opened 2024-06-07
feat: Delete verification records

bhavanakarwade:
### What?
- Implement delete verification records API

### What?
- To delete unnecessary verification records


KulkarniShashank (approved):
LG

PR #760, opened 2024-06-10
refactor: ledger config query function

bhavanakarwade:
### What?
- Refactored the `createLedgerConfig` function in the seed.ts file.

### Why?
- To update the master table `ledgerConfig` with the latest entries.


KulkarniShashank (approved):
LG

PR #769, opened 2024-06-13
fix: incorrect response while creating an offer using multiple connection Ids

bhavanakarwade:
### What?
- Resolved below bug:
 1). An incorrect response occurs while creating an issuance offer (https://github.com/credebl/automation-test-suite/issues/431)

### Why/
- To provide clear feedback on the status of each issuance attempt, improving error handling and user experience


GHkrishna (approved):
LG

PR #772, opened 2024-06-14
fix: improve exception handling

bhavanakarwade:
### What?
- Refactored global exception handler function


KulkarniShashank (approved):
LG

PR #789, opened 2024-06-18
fix: added validation for network and ledgerId update when did update

KulkarniShashank:
### What ###

- Implemented validation logic to ensure network and ledgerId are correctly updated when a DID is updated.

### Why ###

- Previously, there was no validation in place for network and ledgerId during a DID update. This could lead to inconsistencies and potential issues with the integrity of the data.

### How ###

- Added validation checks in the update function for DIDs.
- Modified the update process to include conditions that check the validity of the network and ledgerId.
- Updated relevant tests to cover scenarios where the network and ledgerId are updated to ensure they pass the new validation rules.


GHkrishna (approved):
LG

tipusinghaw (approved):

PR #791, opened 2024-06-19
fix: enhance wallet deletion validation and add null check for orgId in schema and credential-definition

KulkarniShashank:
### What ###

- Improved the validation logic for deleting a wallet, ensuring that references in related tables are properly checked.
- Added validation to ensure orgId is null or not in schema and credential definition entities.

### Why ###

- To prevent accidental deletion of wallets that are still referenced in other parts of the database, which could lead to data integrity issues.
- To ensure that orgId is always present in schema and credential definition entities, null reference as an optional.

### How ###

- Implemented additional validation checks before allowing the deletion of a wallet, ensuring no existing references are left in related tables.
- Added validation rules in the schema and credential definition code to enforce that orgId null as an optional


bhavanakarwade (approved):

GHkrishna (approved):
LG

PR #797, opened 2024-06-20
fix: add condition to check ledgerId is not null

GHkrishna:
## What
- Added the validation for create DID when ledgerId is null
## Why
- When ledgerId is null, DID does not create with a specific ledger


pranalidhanavade (commented):
LG

pranalidhanavade (approved):

PR #813, opened 2024-06-26
refactor: get all schemas API

bhavanakarwade:
### What?
- Refactored GET all platform schemas API (provide support for filtering schemas by schemaType)


GHkrishna (approved):
LG

PR #814, opened 2024-06-26
fix: test file upload on develop

tipusinghaw:
#### What
- Added logger to check file upload flow


GHkrishna (approved):
LG,
Will be reverted once tested

PR #822, opened 2024-07-02
fix: create organisation changes and added geo-location attribute in response.

ganeshawle25:
### What ###

- Added geo-location attribute in create-organisation-dto class.
- Select those attributes in create organisation response and in user-profile response.

### Why ###

- Need to store while creating organisation.


shitrerohit (approved):

pranalidhanavade (approved):
LG

bhavanakarwade (approved):

PR #823, opened 2024-07-02
fix: send email verification issue

bhavanakarwade:
### What?
- Resolved below issue:
When an error occurs while sending the verification email and the same issue persists upon retrying, the response states that the verification email has already been sent.

### Why?
- To prevent incorrect state and misleading error messages


tipusinghaw (approved):

GHkrishna (approved):
LG

PR #824, opened 2024-07-03
refactor: bulk issuance template list response changes

tipusinghaw:
#### What
- Removed logic to check schema already exists for W3C
- Changed template list response
- Added validation for empty attribute


GHkrishna:
Why are we removing logic for checking if schema exists for w3c?

GHkrishna (approved):

PR #837, opened 2024-07-11
fix: parameter and response related bugs

bhavanakarwade:
### What?
- Worked on automation board bugs

1. https://github.com/credebl/automation-test-suite/issues/410
2. https://github.com/credebl/automation-test-suite/issues/411
3. https://github.com/credebl/automation-test-suite/issues/413
4. https://github.com/credebl/automation-test-suite/issues/409

- Resolved create organization issue (while creating an organization if orgname is ORG, then  the incorrect validations are applied)
- Resolved pagination issue for the GET file-data API


libs/common/src/common.utils.ts
```diff
@@ -7,17 +7,19 @@ export function paginator<T>(
   current_page: number,
   items_per_page: number
 ) {
-  const page = current_page || 1,
-    per_page = items_per_page || 10,
-    offset = (page - 1) * per_page,
-    paginatedItems = items.slice(offset).slice(0, items_per_page),
-    total_pages = Math.ceil(items.length / per_page);
+  const page: number = Number(current_page) || 1;
```
GHkrishna, line 10:
```suggestion
  const page = current_page || 1;
```
won't this work too? because current_page is already set as a number above?

libs/common/src/common.utils.ts
```diff
@@ -7,17 +7,19 @@ export function paginator<T>(
...
+  const per_page: number = Number(items_per_page) || 10;
```
GHkrishna, line 11: similar to above

GHkrishna (approved):
LGTM

libs/common/src/common.utils.ts
bhavanakarwade: This is used to avoid string concatanation

libs/common/src/common.utils.ts
bhavanakarwade: This is used to avoid string concatanation

PR #840, opened 2024-07-11
fix: imageurl issue for verification process

bhavanakarwade:
### What?
- Resolved imageUrl issue for verification method


pallavighule (approved):
LGTM

PR #851, opened 2024-07-17
feat: added holder role guard

KulkarniShashank:
### What ###

- Added validation for the organization role guard and user access guard.
- Updated the user decorator for user roles.

### Why ###

- To ensure that only users with the appropriate roles and permissions can access certain functionalities.
- To enhance the security and integrity of the application by enforcing role-based access control.

### How ###

- Implemented new validation logic for the organization role guard.
- Added checks for user access guard to ensure proper user validation.
- Modified the user decorator to handle user roles more effectively.

### DIFF
--- a/apps/api-gateway/src/user/dto/add-user.dto.ts
+++ b/apps/api-gateway/src/user/dto/add-user.dto.ts
@@ -38,7 +38,7 @@ export class AddUserDetailsDto {
 
     @ApiPropertyOptional({ example: false })
     @IsOptional()
-    @IsBoolean({ message: 'holder should be boolean' })
+    @IsBoolean({ message: 'isHolder should be boolean' })
     isHolder?: boolean;
 }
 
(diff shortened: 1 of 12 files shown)


apps/api-gateway/src/user/dto/add-user.dto.ts
```diff
@@ -35,6 +35,11 @@ export class AddUserDetailsDto {
     @IsOptional()
     @IsBoolean({ message: 'isPasskey should be boolean' })
     isPasskey?: boolean;
+
+    @ApiPropertyOptional({ example: false })
+    @IsOptional()
+    @IsBoolean({ message: 'holder should be boolean' })
```
tipusinghaw, line 41:
```suggestion
    @IsBoolean({ message: 'isHolder should be boolean' })
```

tipusinghaw (approved):

PR #857, opened 2024-07-18
refactor: store holder role in user role mapping table

KulkarniShashank:
### What ###
Refactor the code to store the holder role in the user role mapping table.

### Why ###
Storing the holder role in a dedicated user role mapping table enhances maintainability, scalability, and clarity in role management.

### How ###

- Created a new user role mapping table to store user roles.
- Updated the existing code to insert and retrieve holder roles from the new table.
- Modified related functions and queries to accommodate the new structure.
- Added necessary migrations to update the database schema.


tipusinghaw (approved):
LGTM

PR #858, opened 2024-07-18
feat: configure cloud base wallet

KulkarniShashank:
### What ###
Implement an API for configuring a cloud-based wallet.

### Why ###
This feature enables admin to securely manage and store their wallets stored in the database

### How ###
Create API for base wallet configure and manage in `cloud_wallet_user_info` table


tipusinghaw (approved):
LGTM

PR #878, opened 2024-07-29
hotfix: Store connection id in presentation table

bhavanakarwade:
### What?
- Store connection id in presnetation table


pallavighule (approved):
LG

PR #894, opened 2024-08-05
fix: gracefully handle missing deeplink domain url from env

GHkrishna:
# What:
- Gracefully handle missing environment variable for deeplink domain url

# Why:
- Currently with missing env variable, the exception results in unexpected errors


bhavanakarwade (approved):

KulkarniShashank (approved):
LG

PR #896, opened 2024-08-05
fix: added user email validation for holder role guard

KulkarniShashank:
### What ###
Implemented email validation for users with the "holder" role in jwt strategy file.

### Why ###
Ensures that only valid email addresses are used for check the email exists or not in token, improving data integrity and preventing potential errors related to incorrect email formats.

### How ###
Added validation logic in the role guard to check the email format for users assigned the holder role.


GHkrishna (approved):
LG

PR #899, opened 2024-08-07
hotfix: create organization issue

bhavanakarwade:
### What?
- Resolved create organization issue (while creating an organization if orgname is ORG, then the incorrect validations are applied)


GHkrishna:
This change is already being pushed from dev to qa and hence will be deployed on Main with the latest release. Closing as not required.

PR #906, opened 2024-08-09
fix: added-credential-data-in-credential-webhook

pallavighule:
### What ###

- Added credential formatdata in credential webhook

### Why ###

- To get credential format data in credential webhook


GHkrishna (approved):
LG

PR #916, opened 2024-08-16
refactor: get all proofs function

bhavanakarwade:
### What?
- Refactor fetch all proof details function to fetch old records


pallavighule (approved):
LGTM

PR #957, opened 2024-08-29
fix: schema search during pagination

GHkrishna:
## What:
- Fixes schema search due to default pageNumber
## Why:
- Previously search text only allowed searching schema from 1st paginated page (by default). This caused issue for searching schema across all available schemas


tipusinghaw (approved):

pranalidhanavade (approved):
LGTM

PR #962, opened 2024-08-30
fix: schema name is not getting in credentials list

bhavanakarwade:
### What?
- Resolved schema id is not getting in credentials list for oob and bulk issuance issue


GHkrishna (approved):
LG

PR #975, opened 2024-09-03
feat: get-all-file-details-and-file-data

pallavighule:
### What ###

- Get all file details and file data

### DIFF
--- a/apps/api-gateway/src/issuance/issuance.controller.ts
+++ b/apps/api-gateway/src/issuance/issuance.controller.ts
@@ -508,15 +508,15 @@ async downloadBulkIssuanceCSVTemplate(
     return res.status(HttpStatus.OK).json(finalResponse);
   }
   @Get('/orgs/:orgId/:fileId/bulk/file-details-and-file-data')
-  @Roles(OrgRoles.OWNER, OrgRoles.ADMIN, OrgRoles.ISSUER, OrgRoles.VERIFIER)
+  @Roles(OrgRoles.ADMIN, OrgRoles.VERIFIER, OrgRoles.ISSUER, OrgRoles.OWNER)
   @UseGuards(AuthGuard('jwt'), OrgRolesGuard)
   @ApiBearerAuth()
-  @ApiResponse({ status: HttpStatus.OK, description: 'Success', type: ApiResponseDto })
   @ApiUnauthorizedResponse({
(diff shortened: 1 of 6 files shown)


apps/api-gateway/src/issuance/issuance.service.ts
```diff
@@ -162,5 +162,12 @@ export class IssuanceService extends BaseService {
         const payload = { orgId, userDetails };
         return this.sendNatsMessage(this.issuanceProxy, 'delete-issuance-records', payload);
     }
+    async getFileDetailsAndFileDataByFileId(orgId: string, fileId: string): Promise<{ response: object }> {
+        const payload = {
+          orgId,
+          fileId
+        };
+        return this.sendNats(this.issuanceProxy, 'issued-file-data-and-file-details', payload);
```
bhavanakarwade, line 170: Recently, we've been using the `sendNatsMessage` function. In this function, there's no need to add response as a key, you can use Promise<object> instead. Please refer to the existing functions for reference.

apps/api-gateway/src/issuance/issuance.controller.ts
```diff
@@ -507,6 +507,39 @@ async downloadBulkIssuanceCSVTemplate(
...
+    @Param('orgId') orgId: string,
+    @Param(new ValidationPipe({ transform: true })) query: FileQuery,
+    @Res() res: Response
+  ): Promise<object> {
+    const { fileId } = query;
+    const issuedFileDetails = await this.issueCredentialService.getFileDetailsAndFileDataByFileId(orgId, fileId);
+    const finalResponse: IResponseType = {
+      statusCode: HttpStatus.OK,
+      message: ResponseMessages.issuance.success.fileDetailsAndFileData,
+      data: issuedFileDetails.response
```
bhavanakarwade, line 539: Using `sendNatsMessage` which I have mentioned in service function, you can simply use data: issuedFileDetails. As part of the new API refactorization task, we no longer include the response key in the data. Please refer to other functions for reference

bhavanakarwade (changes requested):

apps/api-gateway/src/issuance/issuance.service.ts
pallavighule: resolved

bhavanakarwade (approved):
LGTM

PR #994, opened 2024-09-16
fix: validations

bhavanakarwade:
### What?
- Applied trim and empty string validations in get verified proof details API parameters


GHkrishna (approved):
LG

KulkarniShashank (approved):

PR #997, opened 2024-09-18
fix: added-file-upload-id-in-socket-emitting events

pallavighule:
### What ###

- Added file upload id in socket emitting events


tipusinghaw (approved):
LGTM

PR #1003, opened 2024-09-24
fix: provision for parentThreadId in proofRequest

RinkalBhojani:
what?
Added provision to pass parentThreadId in proofRequest creation payload and use it in further process

why?
As a part of NDI requirement we needed this to pass from proof request creation payload. We did not had this provision in CREDEBL, hence adding this PR.


tipusinghaw (approved):

RinkalBhojani:
Closing this PR because I have manually set DCO to pass. I will create new PR to correct this

PR #1020, opened 2024-10-07
fix: auto start services with docker

GHkrishna:
## What:
- Add minimum changeable .env.sample for Platform and agent for starting services effortlessly (Especially, for provisioning the agent in the agent-provisioning service)
- Initiating service with Docker with the docker-compose up file. This PR ensures that the platform will be able to start using docker
- Add new services that have not been included in the dockerfile


tipusinghaw (approved):
LGTM

bhavanakarwade (approved):
LG

PR #1039, opened 2024-10-18
refactor: added height and width in prettyVc object

bhavanakarwade:
### What?
- Added height and width configurations in prettyVc object for bulk issuance and retry bulk issuance

### DIFF
--- a/apps/issuance/src/issuance.service.ts
+++ b/apps/issuance/src/issuance.service.ts
@@ -1531,7 +1531,9 @@ async sendEmailForCredentialOffer(sendEmailCredentialOffer: SendEmailCredentialO
           platformName: clientDetails?.platformName,
           certificate: clientDetails?.certificate,
           size: clientDetails?.size,
-          orientation: clientDetails?.orientation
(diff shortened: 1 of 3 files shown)


pallavighule (approved):
LG

apps/issuance/src/issuance.service.ts
```diff
@@ -1366,7 +1366,9 @@ async sendEmailForCredentialOffer(sendEmailCredentialOffer: SendEmailCredentialO
           platformName: bulkPayloadDetails?.platformName,
           certificate: bulkPayloadDetails?.certificate,
           size: bulkPayloadDetails?.size,
-          orientation: bulkPayloadDetails?.orientation
+          orientation: bulkPayloadDetails?.orientation,
+          height: bulkPayloadDetails?.height,
```
tipusinghaw, line 1370: `bulkPayloadDetails` already destructured at the starting of the function, you can use the value in similar way

tipusinghaw (approved):

PR #1050, opened 2024-10-23
chore: transition from Blockster Labs to Ayanworks

GHkrishna:
## What
- Update occurrences of `Blockster Labs` to `Ayanworks`

### DIFF
--- a/.github/ISSUE_TEMPLATE/DMP_2024.yml
+++ b/.github/ISSUE_TEMPLATE/DMP_2024.yml
@@ -76,7 +76,7 @@ body:
       multiple: false
       options:
         - Bandhu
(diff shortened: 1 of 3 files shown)


.github/ISSUE_TEMPLATE/DMP_2024.yml
```diff
@@ -76,7 +76,7 @@ body:
       multiple: false
       options:
         - Bandhu
-        - Blockster Labs (CREDEBL)
```
ajile-in, line 79: Same, no change needed in this file. Keep as it is.

ajile-in (changes requested):
We need to keep Blockster Labs name in this file.

PR #1055, opened 2024-10-24
refactor:  made `isEcosystemEnabled` flag configurable

bhavanakarwade:
### What?
- Made `isEcosystemEnabled` flag configurable


apps/organization/src/organization.service.ts
```diff
@@ -623,26 +623,38 @@ export class OrganizationService {
         userId
       );
 
-      let orgIds;
-      if (0 > getOrgs?.organizations?.length) {
-        orgIds = getOrgs?.organizations?.map(item => item.id);
+      if (0 === getOrgs?.organizations?.length) {
```
tipusinghaw, line 626:
```suggestion
      const {organizations} = getOrgs
      if (0 === getOrgs?.organizations?.length) {
```

apps/organization/src/organization.service.ts
```diff
@@ -623,26 +623,38 @@ export class OrganizationService {
...
       }
-       
-      const orgEcosystemDetails = await this._getOrgEcosystems(orgIds);
-      if (!orgEcosystemDetails || !Array.isArray(orgEcosystemDetails) || 0 === orgEcosystemDetails.length) {
-        throw new NotFoundException(ResponseMessages.ecosystem.error.ecosystemDetailsNotFound);
+
+      let orgIds;
+      let updatedOrgs;
+
+      if ('true' === process.env.IS_ECOSYSTEM_ENABLE) {
```
GHkrishna, line 633: Can we please add this with the appropriate value in the sample.env and demo.env and in our env trancking

libs/common/src/response-messages/index.ts
```diff
@@ -121,6 +121,7 @@ export const ResponseMessages = {
       invalidClient: 'Invalid client credentials',
       primaryDid: 'This DID is already set to primary DID',
       didNotFound: 'DID does not exist in organiation',
+      organizationNotFound: 'No any organizations found',
```
tipusinghaw, line 124: Please change this message

apps/organization/src/organization.service.ts
```diff
@@ -623,26 +623,38 @@ export class OrganizationService {
...
+        if (0 > getOrgs?.organizations?.length) {
```
GHkrishna, line 634: Is this if condition required, since we already check the length above to throw error
`      if (0 === getOrgs?.organizations?.length) {`

apps/organization/src/organization.service.ts
```diff
@@ -623,26 +623,38 @@ export class OrganizationService {
...
+        updatedOrgs = getOrgs.organizations.map(org => {
+          const matchingEcosystems = orgEcosystemDetails
+            .filter(ecosystem => ecosystem.orgId === org.id)
+            .map(ecosystem => ({ ecosystemId: ecosystem.ecosystemId }));
+          return {
+            ...org,
+            ecosystemOrgs: 0 < matchingEcosystems.length ? matchingEcosystems : []
+          };
+        });
+      } else {
```
GHkrishna, line 652: This `else` is for org without any ecosystem associated, right?

GHkrishna (changes requested):

apps/organization/src/organization.service.ts
bhavanakarwade: Resolved.

libs/common/src/response-messages/index.ts
bhavanakarwade: Resolved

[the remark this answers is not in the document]
bhavanakarwade: Resolved

apps/organization/src/organization.service.ts
bhavanakarwade: I have added this variable in sample.env and demo.env

apps/organization/src/organization.service.ts
bhavanakarwade: Yes, Correct

apps/organization/src/organization.service.ts
bhavanakarwade: Yes, Correct. We have already checking the length, So there is no need to add `if (0 > getOrgs?.organizations?.length)`, this condition again.

.env.demo
```diff
@@ -128,3 +128,5 @@ AGENT_PROTOCOL=http
 OOB_BATCH_SIZE=10
 MAX_ORG_LIMIT=10
 FIDO_API_ENDPOINT=http://localhost:8000
+
+IS_ECOSYSTEM_ENABLE=true
```
GHkrishna, line 132: Can you make this as false as we won't be needing to have ecosystem enables

.env.sample
```diff
@@ -153,4 +153,6 @@ SCHEMA_FILE_SERVER_TOKEN=xxxxxxxx // Please provide schema file server token for
 
 FILEUPLOAD_CACHE_TTL= //Provide file upload cache ttl
 
-FIELD_UPLOAD_SIZE= //Provide field upload size
\ No newline at end of file
+FIELD_UPLOAD_SIZE= //Provide field upload size
```
GHkrishna, line 156: Is this missing in `.env.demo`

GHkrishna (changes requested):

.env.demo
bhavanakarwade: Resolved

GHkrishna (approved):

PR #1078, opened 2024-11-15
feat: archive schema for deleted wallet

bhavanakarwade:
### What?
- Implemented schema archiving functionality for deleted wallets.

### DIFF
--- a/apps/ledger/src/schema/repositories/schema.repository.ts
+++ b/apps/ledger/src/schema/repositories/schema.repository.ts
@@ -179,7 +179,7 @@ export class SchemaRepository {
     }
   }
 
-  async updateSchemaDetails(did: string): Promise<Prisma.BatchPayload> {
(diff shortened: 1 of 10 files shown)


GHkrishna (approved):

apps/ledger/src/schema/repositories/schema.repository.ts
```diff
@@ -176,6 +179,24 @@ export class SchemaRepository {
     }
   }
 
+  async updateSchemaDetails(did: string): Promise<Prisma.BatchPayload> {
```
GHkrishna, line 182: We can maybe rename this function to something like `archieveSchema` or `archieveSchemaByDid`.
What do you think?

bhavanakarwade: Yes, correct. I think archiveSchema is a more descriptive function name.

GHkrishna (approved):

PR #1086, opened 2024-12-05
feat: Add support to select multiple connections while sending proof request

pranalidhanavade:
### What

- Added support for multiple connections while sending proof request.
- Done DTO changes for indy and jsonld.

### DIFF
--- a/apps/api-gateway/src/verification/dto/request-proof.dto.ts
+++ b/apps/api-gateway/src/verification/dto/request-proof.dto.ts
@@ -141,7 +141,7 @@ export class ProofRequestPresentationDefinition {
   @IsOptional()
   name: string;
 
-  @ApiProperty()
+  @ApiProperty({type: () => [InputDescriptors]})
   @IsArray({ message: 'inputDescriptors must be an array' })
   @IsNotEmpty({ message: 'inputDescriptors is required.' })
   @ArrayMinSize(1)
@@ -274,14 +274,14 @@ export class RequestProofDto extends ProofPayload {
 
     orgId: string;
 
-    @ApiPropertyOptional()
+    @ApiPropertyOptional({enum:AutoAccept})
     @IsString({ message: 'auto accept proof must be in string' })
     @IsNotEmpty({ message: 'please provide valid auto accept proof' })
     @IsOptional()
     @IsEnum(AutoAccept, {
         message: `Invalid auto accept proof. It should be one of: ${Object.values(AutoAccept).join(', ')}`
     })
-    autoAcceptProof: string;
+    autoAcceptProof: AutoAccept;
 }
 
 export class OutOfBandRequestProof extends ProofPayload {
--- a/apps/api-gateway/src/verification/dto/request-proof.dto.ts
+++ b/apps/api-gateway/src/verification/dto/request-proof.dto.ts
@@ -1,4 +1,4 @@
-import { ArrayNotEmpty, IsArray, IsBoolean, IsEmail, IsEnum, IsNotEmpty, IsNumberString, IsObject, IsOptional, IsString, ValidateIf, ValidateNested, IsUUID, ArrayUnique, ArrayMaxSize } from 'class-validator';
+import { ArrayNotEmpty, IsArray, IsBoolean, IsEmail, IsEnum, IsNotEmpty, IsNumberString, IsObject, IsOptional, IsString, ValidateIf, ValidateNested, IsUUID, ArrayUnique, ArrayMaxSize, ArrayMinSize } from 'class-validator';
 import { trim } from '@credebl/common/cast.helper';
 import { ApiProperty, ApiPropertyOptional } from '@nestjs/swagger';
 import { Transform, Type } from 'class-transformer';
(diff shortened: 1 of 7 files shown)


tipusinghaw (approved):

apps/api-gateway/src/verification/dto/request-proof.dto.ts
```diff
@@ -192,12 +195,17 @@ export class IndyDto {
...
+    example: ['32f54163-7166-48f1-93d8-ff217bdb0653']
+  })  
+
+    @IsNotEmpty({ each: true, message: 'connectionId array elements must not be empty.' })
+    @ValidateIf((obj) => Array.isArray(obj.connectionId))
+    @IsArray({ message: 'connectionId must be an array.' })
+    @ArrayMinSize(1, { message: 'connectionId must contain at least 1 element.' })
+    @ArrayMaxSize(10, { message: 'connectionId can contain at most 10 elements.' })
+    @IsUUID('all', { each: true, message: 'Each connectionId must be a valid UUID.' })
+    connectionId: string | string[];
```
bhavanakarwade, line 208: property name should be descriptive. Because we are using connection Id's array here.

apps/verification/src/verification.service.ts
```diff
@@ -227,64 +227,74 @@ export class VerificationService {
...
+      const connectionIds = Array.isArray(requestProof.connectionId)
+        ? requestProof.connectionId
+        : [requestProof.connectionId];
+  
+      for (const connectionId of connectionIds) {
+        const proofRequestPayload = {
+          comment,
+          connectionId,
+          autoAcceptProof: requestProof.autoAcceptProof || AutoAccept.Never,
+          goalCode: requestProof.goalCode || undefined,
```
bhavanakarwade, line 253: Use object destructuring here.

apps/api-gateway/src/verification/dto/request-proof.dto.ts
pranalidhanavade: I reviewed some best practices for naming DTOs and came across the following references for such scenarios.

`expectedRPID?: string | string[];`

apps/verification/src/verification.service.ts
pranalidhanavade: Resolved

apps/api-gateway/src/verification/dto/request-proof.dto.ts
```diff
@@ -138,12 +141,12 @@ export class ProofRequestPresentationDefinition {
   @IsOptional()
   name: string;
 
-  @ApiProperty({type: () =>  [InputDescriptors]})
```
GHkrishna, line 141: Any specific reason to remove this?

apps/api-gateway/src/verification/dto/request-proof.dto.ts
```diff
@@ -272,7 +281,7 @@ export class RequestProofDto extends ProofPayload {
     @IsEnum(AutoAccept, {
         message: `Invalid auto accept proof. It should be one of: ${Object.values(AutoAccept).join(', ')}`
     })
-    autoAcceptProof: AutoAccept;
```
GHkrishna, line 275: Why have we removed this?

GHkrishna (changes requested):

apps/api-gateway/src/verification/dto/request-proof.dto.ts
GHkrishna, line 208: can we simply have connectionId as an array of string. So in cases where we only have one `connectionId` we enter,
`
connectionId: ["some-uuid"]
`

apps/verification/src/interfaces/verification.interface.ts
```diff
@@ -31,6 +31,45 @@ export interface IRequestProof {
...
+export interface IProofAttributesData {
+    attributeName: string;
+    attributeNames?: string[];
+    condition: string;
+    value: string;
+    credDefId: string;
+    schemaId: string;
+}
+
+export interface ProofRequestAttributeDto {
```
GHkrishna, line 65: I dont see this interface being used anywhere

apps/verification/src/verification.service.ts
```diff
@@ -627,16 +648,15 @@ export class VerificationService {
               p_value: parseInt(attribute.value),
               restrictions: [
                 {
-                  cred_def_id: proofRequestpayload.attributes[index].credDefId ? proofRequestpayload.attributes[index].credDefId : undefined,
-                  schema_id: proofRequestpayload.attributes[index].schemaId
+                  cred_def_id: proofRequestpayload?.proofFormats?.indy?.attributes[index].credDefId ? proofRequestpayload?.proofFormats?.indy?.attributes[index].credDefId : undefined,
```
GHkrishna, line 651: in case the `attributes[index].credDefId: undefined`, does the proofrequest allow any credDef or does the credDef only needs to be `undefined`?

GHkrishna (changes requested):

apps/api-gateway/src/verification/dto/request-proof.dto.ts
pranalidhanavade: Yes, We can.

apps/api-gateway/src/verification/dto/request-proof.dto.ts
pranalidhanavade: Resolved

apps/verification/src/verification.service.ts
pranalidhanavade: No it doesn't allow credDef in this case `attributes[index].credDefId: undefined`.

PR #1088, opened 2024-12-18
fix: add declined column in proof webhook

tipusinghaw:
#### What
- Stored decline message for proof webhook


pranalidhanavade (approved):
LG

GHkrishna (approved):
LG

PR #1091, opened 2024-12-30
feat: global NATS interceptor implementation

tipusinghaw:
#### What
- Implemented NATS interceptor for NATS error message
#### Why
- Not able to customize the NATS error messages

### DIFF
--- a/apps/api-gateway/common/exception-handler.ts
+++ b/apps/api-gateway/common/exception-handler.ts
@@ -33,7 +33,7 @@ export class CustomExceptionFilter extends BaseExceptionFilter {
     if ('string' === typeof exceptionResponseData) {
       exceptionResponse.message = exceptionResponseData;
     } else {
-      exceptionResponse = exceptionResponseData as ExceptionResponse;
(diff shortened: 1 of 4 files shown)


GHkrishna:
Good job :+1: 

Will we be also getting the service name which throws the error in the error message.

GHkrishna (approved):
LG

tipusinghaw:
> Good job 👍
> 
> Out of curiosity, will we also be getting the service name which throws the error in the error message.



> Not in the current state of the codebase, for that we might need to intercept the message from every service and add the service name (Just a thought)

GHkrishna:
> Not in the current state of the codebase, for that we might need to intercept the message from every service and add the service name (Just a thought)

That seems interesting, it would be great if we could achieve that. Because that way we can easily recognize the service that might be down.

We can add an interceptor like you mentioned, and maybe explore if `NATS` provides any out of the box capabilities or if we can provision service name in the error object at the time of throwing the error itself

ankita-p17 (approved):

PR #1111, opened 2025-02-13
fix: added search on schema name for credentials list

bhavanakarwade:
### What?
- Added search on schema name for credentials list

### DIFF
--- a/apps/issuance/src/issuance.service.ts
+++ b/apps/issuance/src/issuance.service.ts
@@ -461,8 +461,11 @@ export class IssuanceService {
       let schemaIds;
       if (issuedCredentialsSearchCriteria?.search) {
         const schemaDetails = await this._getSchemaDetailsByName(issuedCredentialsSearchCriteria?.search);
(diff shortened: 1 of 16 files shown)


apps/issuance/src/issuance.service.ts
```diff
@@ -457,10 +457,18 @@ export class IssuanceService {
     issuedCredentialsSearchCriteria: IIssuedCredentialSearchParams
   ): Promise<IIssuedCredential> {
     try {
+
+      let schemaIds;
+      if (issuedCredentialsSearchCriteria?.search) {
+        const schemaDetails = await this._getSchemaDetailsByName(issuedCredentialsSearchCriteria?.search);
+        schemaIds = schemaDetails?.map(item => item?.schemaLedgerId);
```
tipusinghaw, line 464: This might break if the `schemaDetails` is empty

bhavanakarwade: Yes, I have handled the conditions if array length comes 0.

tipusinghaw (approved):

PR #1124, opened 2025-03-03
fix: added API params validations

pranalidhanavade:
### What 

- Fix: added API params validation for orgId in agent health API.


GHkrishna (approved):
LG

tipusinghaw (approved):

PR #1139, opened 2025-03-20
fix:add tenant id in get org info

pallavighule:
### What ###

- Add tenant id in get org info API

### Why ###

- To get tenantId in the Get Org Info API response


tipusinghaw (approved):

GHkrishna (approved):
LG

PR #1155, opened 2025-03-27
fix: local setup issues

GHkrishna:
## What:
- Some of the latest docker version requires `docker compose` instead of previously used `docker-compose`
- Change in scripts for local setup in the `.demo.env`. This configurations are only enabled for local setup using `pnpm` and not for `docker`ized builds

### DIFF
--- a/apps/agent-provisioning/AFJ/scripts/start_agent.sh
+++ b/apps/agent-provisioning/AFJ/scripts/start_agent.sh
@@ -209,7 +209,7 @@ if [ $? -eq 0 ]; then
 
   docker rm -f "${PROJECT_NAME}" || true
 
-  docker-compose -f $FILE_NAME --project-name "${PROJECT_NAME}" up -d
+  docker compose -f $FILE_NAME --project-name "${PROJECT_NAME}" up -d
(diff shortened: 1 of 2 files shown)


apps/agent-provisioning/AFJ/scripts/start_agent.sh
```diff
@@ -209,7 +209,7 @@ if [ $? -eq 0 ]; then
 
   docker rm -f "${PROJECT_NAME}" || true
 
-  docker-compose -f $FILE_NAME --project-name "${PROJECT_NAME}" up -d
+  docker compose -f $FILE_NAME --project-name "${PROJECT_NAME}" up -d
```
sairanjit, line 212: Is this backward compatible ?

GHkrishna: As per the docs: https://docs.docker.com/compose/install/
The docker-compose is deprecated as it can to endoflife.

Here a supporting doc: https://forums.docker.com/t/docker-compose-vs-docker-compose/137884/6

KambleSahil3 (approved):

sairanjit (approved):

PR #1164, opened 2025-04-02
refactor: remove image-service from libs

sairanjit:
### Summary

- Remove image-service from libs
- Update `lint-staged` to use `prettier` and `eslint`


tipusinghaw (approved):
LGTM

PR #1166, opened 2025-04-03
feat: Support nested attributes while creating schema

bhavanakarwade:
- [x] Support nested attributes while creating schema
- [x] Modified json builder function for migrating draft-07 version to draft 2020-12 version
- [x] Modified download csv functionality for supporting nested structure
- [x] Modified upload csv functionality for supporting nested structure


apps/api-gateway/src/issuance/issuance.service.ts
```diff
@@ -35,7 +35,7 @@ export class IssuanceService extends BaseService {
...
+            payload = { credential: issueCredentialDto.credential, options: issueCredentialDto.options, comment: issueCredentialDto.comment, orgId: issueCredentialDto.orgId, protocolVersion: issueCredentialDto.protocolVersion, goalCode: issueCredentialDto.goalCode, parentThreadId: issueCredentialDto.parentThreadId, willConfirm: issueCredentialDto.willConfirm, label: issueCredentialDto.label, autoAcceptCredential: issueCredentialDto.autoAcceptCredential, credentialType: issueCredentialDto.credentialType, isShortenUrl: issueCredentialDto.isShortenUrl, reuseConnection : issueCredentialDto.reuseConnection, isValidateSchema: issueCredentialDto.isValidateSchema};
```
GHkrishna, line 38: I'm unsure if creating an object this was is a good approach. Can we de-structure the object instead to remove the unwanted additional properties for both the credential types?

libs/validations/arrayKeyword.ts
GHkrishna, line 1: I think instead of creating new file for each validations, it makes sense to have all the validations in a validations.ts file.
Let me know what do you think?

GHkrishna (changes requested):
Great job Bhavana.
Have left some comments.

apps/api-gateway/src/issuance/issuance.service.ts
bhavanakarwade: Yes, we can destructure the object. I have implemented requested suggestions.

apps/ledger/src/schema/enum/schema.enum.ts
```diff
@@ -11,5 +11,5 @@ export enum SortFields {
 }
 
 export enum W3CSchemaVersion {
-    W3C_SCHEMA_VERSION = 'draft-07'
+    W3C_SCHEMA_VERSION = 'draft/2020-12'
```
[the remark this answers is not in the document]
bhavanakarwade: As per discussion with team, We have decided to remove draft-07 version, but user can still issue and verify by existing schemas on draft-07 version.

libs/validations/arrayKeyword.ts
bhavanakarwade: To handle specific validations while building JSON schema, such as enforcing constraints like minLength and maxLength for string fields. For that we referred to the jsonbuilder NPM package. This package includes support for validating various data types across multiple files, making it easier to implement field-specific validation rules during schema generation.

GHkrishna (approved):

PR #1187, opened 2025-04-22
[updated] Add resend verification email functionality with retry limit

harshitghagre:


apps/verification/src/verification.service.ts
GHkrishna, line 1: As per the description we wanted to update the existing `send-verification-link` API. 
So this is not the correct place for the changes, please have a look at the appropriate API.

Let me know if you have any doubts

GHkrishna (changes requested):

GHkrishna:
Seems like the requirements were misunderstood for the issue. Closing the PR for now, you can raise a fresh PR with appropriate changes

PR #1188, opened 2025-04-22
Add resend functionality for verification emails in AuthzController 

harshitghagre:


harshitghagre:
@GHkrishna check this PR

apps/verification/src/verification.service.ts
GHkrishna, line 1: This is not the expected file changes, you would want to update [apps/user/src/user.service.ts](https://github.com/credebl/platform/blob/c36950de601147cd87385427c38f564e316853d6/apps/user/src/user.service.ts#L86)

You can have a look at that file

GHkrishna (changes requested):
Can you please have a look at the flow from controller, since the verification changes that you are doing are for verification of credential which is not authentication flow

harshitghagre:
> Can you please have a look at the flow from controller, since the verification changes that you are doing are for verification of credential which is not authentication flow

okay I'll check it

GHkrishna:
BTW loved your enthusiasm and willingness to contribute, @harshitghagre

harshitghagre:
> BTW loved your enthusiasm and willingness to contribute, @harshitghagre

Thanks @GHkrishna . By profession, I'm not a developer, but I'm trying to fix the issue because I'm getting interested in how the resend mail functionality works,I'm still learning  and working on resolving the issue.

GHkrishna:
> By profession, I'm not a developer, but I'm trying to fix the issue because I'm getting interested in how the resend mail functionality works,I'm still learning and working on resolving the issue.

That's even better, we need more people like you. You're going good. All the best.

harshitghagre:
> > By profession, I'm not a developer, but I'm trying to fix the issue because I'm getting interested in how the resend mail functionality works,I'm still learning and working on resolving the issue.
> 
> That's even better, we need more people like you. You're going good. All the best.

could I know which SMTP server are you using ? is it self host or third party?

harshitghagre:
@GHkrishna I made some changes please check it

apps/user/src/user.service.ts
```diff
@@ -83,82 +83,88 @@ export class UserService {
...
+    redirectUrl: string,
+    clientId: string,
+    brandLogoUrl: string,
+    platformName: string,
+    resend: boolean = false,
+  ): Promise<boolean> {
     try {
-      const { email, brandLogoUrl, platformName, clientId, clientSecret } = userEmailVerification;
-  
-      if ('PROD' === process.env.PLATFORM_PROFILE_MODE) {
```
GHkrishna, line 90: We should not remove this condition

apps/user/src/user.service.ts
```diff
@@ -83,82 +83,88 @@ export class UserService {
...
-  
-      const userDetails = await this.userRepository.checkUserExist(email);
-  
-      if (userDetails) {
-        if (userDetails.isEmailVerified) {
-          throw new ConflictException(ResponseMessages.user.error.exists);
-        } else {
-          throw new ConflictException(ResponseMessages.user.error.verificationAlreadySent);
-        }
+      const maxResendAttempts = platformConfigData[0].maxVerificationEmailResend;
```
GHkrishna, line 100: Lets not take it from here but rather from a config file or env

apps/api-gateway/src/authz/authz.controller.ts
```diff
@@ -66,14 +66,31 @@ export class AuthzController {
...
+  })
+  @ApiOperation({
+    summary: 'Send verification email',
+    description: 'Send verification email to new user',
+  })
+  async create(
+    @Body() userEmailVerification: UserEmailVerificationDto,
+    @Res() res: Response,
+  ): Promise<Response> {
+    const { email, resend = false } = userEmailVerification;
```
GHkrishna, line 82: Not sure why email needs to be de-structured here and how would resend be part of the object, we can instead have a simple resend const declaration if email is not required

apps/api-gateway/src/authz/authz.controller.ts
```diff
@@ -66,14 +66,31 @@ export class AuthzController {
...
+    status: HttpStatus.CREATED,
+    description: 'Created',
+    type: ApiResponseDto,
+  })
+  @ApiOperation({
+    summary: 'Send verification email',
+    description: 'Send verification email to new user',
+  })
+  async create(
+    @Body() userEmailVerification: UserEmailVerificationDto,
```
GHkrishna, line 79: We can add an optional resend flag without any API properties in the `UserEmailVerificationDto`

apps/api-gateway/src/authz/authz.controller.ts
```diff
@@ -66,14 +66,31 @@ export class AuthzController {
...
+    const { email, resend = false } = userEmailVerification;
+
+    // Call the AuthzService method to handle resend logic and retry limit
+    await this.authzService.sendVerificationMail(userEmailVerification, resend);
+
     const finalResponse: IResponseType = {
       statusCode: HttpStatus.CREATED,
-      message: ResponseMessages.user.success.sendVerificationCode
+      message: resend
+        ? ResponseMessages.user.success.resendVerificationCode
```
GHkrishna, line 90: This needs to be added in the `ResponseMessages` with appropriate messages

apps/user/src/user.service.ts
```diff
@@ -83,82 +83,88 @@ export class UserService {
...
+      const platformConfigData = await this.prisma.platform_config.findMany();
+      if (!platformConfigData || platformConfigData.length === 0) {
+        throw new InternalServerErrorException('Platform configuration not found.');
       }
-  
-      const userDetails = await this.userRepository.checkUserExist(email);
-  
-      if (userDetails) {
-        if (userDetails.isEmailVerified) {
-          throw new ConflictException(ResponseMessages.user.error.exists);
```
GHkrishna, line 102: Not sure why this was removed, there can be a case where the user might not exist in the DB and hence we want to add them first, irrespective of the resend flag

apps/user/src/user.service.ts
```diff
@@ -83,82 +83,88 @@ export class UserService {
...
-        } else {
-          throw new ConflictException(ResponseMessages.user.error.verificationAlreadySent);
-        }
+      const maxResendAttempts = platformConfigData[0].maxVerificationEmailResend; 
+      const user = await this.prisma.user.findUnique({ where: { email } });
```
GHkrishna, line 101: As mentioned in a separate comment we would also need to handle scenario where a new user tries to verify their email

libs/prisma-service/prisma/schema.prisma
```diff
@@ -4,36 +4,38 @@ generator client {
...
-  publicProfile          Boolean                  @default(false)
-  password               String?                  @db.VarChar
-  orgInvitations         org_invitations[]
-  user_activities        user_activity[]
-  userDevices            user_devices[]
-  userOrgRoles           user_org_roles[]
-  token                  token[]
-  user_role_mapping      user_role_mapping[]
-  cloud_wallet_user_info cloud_wallet_user_info[]
+  id                        String     @id @default(uuid()) @db.Uuid
```
GHkrishna, line 12: Any specific reason to remove the mapping here?

libs/prisma-service/prisma/schema.prisma
```diff
@@ -4,36 +4,38 @@ generator client {
...
+  createDateTime             DateTime   @default(now()) @db.Timestamptz(6)
+  lastChangedDateTime        DateTime   @default(now()) @db.Timestamptz(6)
+  firstName                  String?    @db.VarChar(500)
+  lastName                   String?    @db.VarChar(500)
+  email                      String?    @unique @db.VarChar(500)
```
GHkrishna, line 17: Any specific reason to remove the mapping here?

apps/user/src/user.service.ts
```diff
@@ -168,37 +174,66 @@ export class UserService {
...
       throw new RpcException(error.response ? error.response : error);
     }
   }
-
-  /**
-   *
-   * @param param email, verification code
-   * @returns Email verification succcess
-   */
-
```
GHkrishna, line 201: Lets preserve this

apps/user/src/user.service.ts
```diff
@@ -168,37 +174,66 @@ export class UserService {
...
+        brandLogoUrl, 
+        platformName
+      );
       const isEmailSent = await sendEmail(emailData);
       if (isEmailSent) {
-        return isEmailSent;
+        return true; 
       } else {
-        throw new InternalServerErrorException(ResponseMessages.user.error.emailSend);
+        throw new InternalServerErrorException("Failed to send verification email.");
```
GHkrishna, line 229:
```suggestion
        throw new InternalServerErrorException(ResponseMessages.user.error.emailSend);```

File: apps/user/src/user.service.ts [hunk 12]
```diff
@@ -168,37 +174,66 @@ export class UserService {
...
+      if (resend) {
+        if (user.emailVerificationAttempts >= maxResendAttempts) {
+          throw new Error("You have exceeded the maximum number of verification email attempts. Please contact support.");
+        }
+        const now = new Date();
+        const lastSentDate = user.lastEmailSentDateTime;
+        if (lastSentDate && (now.getTime() - new Date(lastSentDate).getTime()) < 5 * 60 * 1000) {
+          throw new Error("Please wait before requesting another verification email.");
+        }
+        await this.prisma.user.update({
```
GHkrishna, line 199: Updating here seems risky as we are not yet sure, if the email has been sent yet. Would be good to update it once we are sure the email has bee sent

apps/user/src/user.service.ts
```diff
@@ -168,37 +174,66 @@ export class UserService {
...
-
+      const maxResendAttempts = platformConfigData[0].maxVerificationEmailResend;  
+      const user = await this.prisma.user.findUnique({ where: { email } });
+      if (resend) {
+        if (user.emailVerificationAttempts >= maxResendAttempts) {
+          throw new Error("You have exceeded the maximum number of verification email attempts. Please contact support.");
+        }
+        const now = new Date();
+        const lastSentDate = user.lastEmailSentDateTime;
+        if (lastSentDate && (now.getTime() - new Date(lastSentDate).getTime()) < 5 * 60 * 1000) {
```
GHkrishna, line 196: The cool down time seems like a good idea, we can possibly take the cooldown time in minute from a config itself

apps/user/src/user.service.ts
```diff
@@ -168,37 +174,66 @@ export class UserService {
...
+    resend: boolean = false 
+  ): Promise<boolean> {
     try {
       const platformConfigData = await this.prisma.platform_config.findMany();
-
+      const maxResendAttempts = platformConfigData[0].maxVerificationEmailResend;  
+      const user = await this.prisma.user.findUnique({ where: { email } });
+      if (resend) {
+        if (user.emailVerificationAttempts >= maxResendAttempts) {
+          throw new Error("You have exceeded the maximum number of verification email attempts. Please contact support.");
```
GHkrishna, line 192: Error messages are usually taken from `ResponseMessages`. You can add this message there ans then import it here

GHkrishna (changes requested):

harshitghagre:
@GHkrishna some of the previous commits is mergred in currrent commit like user.service.ts ,  authz.controller.ts files  so it will be better that i'll create new PR in which I do the changes

apps/user/src/user.service.ts
harshitghagre: getting it from agent.env or sample.env

GHkrishna:
> it will be better that i'll create new PR in which I do the changes
Sounds right to me, closing the PR

PR #1190, opened 2025-04-22
fix: Update GET proof record endpoint to return 404 for non-existing records

amrrdev:
Issue Fixed:
I addressed the error handling in the proof record retrieval endpoint where it was returning a 500 status code for non-existing records. The issue was in the verification service's error handling logic.

Changes Made:
1. Modified error handling in verification.service.ts for the getProofPresentationById method
2. Updated the error detection to properly identify "record not found" cases
3. Implemented proper 404 response using NestJS's NotFoundException
4. Maintained the original error message for clarity

Before:
- API returned 500 status code with a generic error
- Error structure was inconsistent with API standards
- Generic "Something went wrong!" message wasn't helpful

After:
- API now returns 404 status code for non-existing records
- Clear error message indicates the specific proofId that wasn't found
- Consistent error structure throughout the API
- Better error handling helps client applications handle missing records appropriately

This change improves the API's error handling consistency and provides better feedback for client applications trying to access non-existing proof records.

### DIFF
--- a/apps/verification/src/verification.service.ts
+++ b/apps/verification/src/verification.service.ts
@@ -187,17 +187,13 @@ export class VerificationService {
       return getProofPresentationById?.response;
     } catch (error) {
       this.logger.error(`[getProofPresentationById] - error in get proof presentation by proofId : ${JSON.stringify(error)}`);
-      const errorStack = error?.response?.error?.reason;
+      const errorMessage = error?.response?.error?.reason || error?.message;
 
-      if (errorStack) {
-        throw new RpcException({
-          message: ResponseMessages.verification.error.proofNotFound,
-          statusCode: error?.response?.status,
-          error: errorStack
-        });
-      } else {
-        throw new RpcException(error.response ? error.response : error);
+      if (errorMessage?.includes('not found')) {
+        throw new NotFoundException(errorMessage);
       }
+
+      throw new RpcException(error.response ? error.response : error);
     }
   }
 


apps/verification/src/verification.service.ts
```diff
@@ -187,17 +187,15 @@ export class VerificationService {
...
-      if (errorStack) {
-        throw new RpcException({
-          message: ResponseMessages.verification.error.proofNotFound,
-          statusCode: error?.response?.status,
-          error: errorStack
-        });
-      } else {
-        throw new RpcException(error.response ? error.response : error);
+      // Handle "record not found" error with 404
+      if (errorMessage?.includes('record with id') && errorMessage?.includes('not found')) {
```
GHkrishna, line 193: I suppose we would also need to handle errors with `error?.response?.error?.reason` which might have error others than we have added conditions for which might not be directly available under `err.response`

GHkrishna (changes requested):

apps/verification/src/verification.service.ts
amrrdev: Hi! Thanks a lot for the feedback.
I’ve updated the logic based on your notes. Here's what I’m currently doing:

1. **If the error clearly indicates a "record not found"**, I return a `NotFoundException`:
```ts
if (errorMessage?.includes('record with id') && errorMessage?.includes('not found')) {
  throw new NotFoundException({
    statusCode: 404,
    message: errorMessage,
    error: 'Not found',
  });
}
```

2. **If the error has a structured shape**, I return it as an `RpcException`:
```ts
if (error?.response) {
  throw new RpcException({
    statusCode: error.response.status || 500,
    message: errorMessage || 'Internal server error',
    error: error.response.error || 'Internal error'
  });
}
```

3. **If the error doesn't follow a known structure**, I fall back to a generic `RpcException`:
```ts
throw new RpcException({
  statusCode: 500,
  message: 'Something went wrong!',
  error: errorMessage || 'Unknown error'
});
```
```suggestion
      if (errorMessage?.includes('record with id') && errorMessage?.includes('not found')) {
```
Let me know if this direction looks better or if you'd like me to change how the fallback is handled.

GHkrishna: This seems almost right, however, in either of the conditions, we don't want to ignore messages from error if it has `error.response` but doesn't have error.response.status.
Similar, to the previous implementation:
```ts
throw new RpcException(error.response ? error.response : error);
```

amrrdev: I've updated the error handling logic based on your note. Here's what I'm doing now:

- I'm extracting the `errorMessage` like this:
```ts
const errorMessage = error?.response?.error?.reason || error?.message;
```
This ensures we capture a meaningful message even if `error.response` exists but doesn't include a structured `error.reason`.

- If the message indicates "record not found", I return a `NotFoundException`:
```ts
if (errorMessage?.includes('record with id') && errorMessage?.includes('not found')) {
  throw new NotFoundException({
    statusCode: 404,
    message: errorMessage,
    error: 'Not found'
  });
}
```

- For all other cases, I preserve the original structure:
```ts
throw new RpcException(error.response ? error.response : error);
```

Let me know if this looks aligned with what you had in mind, or if you'd like any further adjustments

bhavanakarwade: Okay this seems alright to me, we can handle it that way

amrrdev: Hi @GHkrishna and @bhavanakarwade ,

I’ve committed the changes with the updated error handling. Let me know if anything else is needed!

Thanks!

apps/verification/src/verification.service.ts
```diff
@@ -187,17 +187,17 @@ export class VerificationService {
...
-        throw new RpcException({
-          message: ResponseMessages.verification.error.proofNotFound,
-          statusCode: error?.response?.status,
-          error: errorStack
+      if (errorMessage?.includes('record with id') && errorMessage?.includes('not found')) {
+        throw new NotFoundException({
+          statusCode: 404,
+          message: errorMessage,
+          error: 'Not found'
         });
```
GHkrishna, line 197:
```suggestion
        throw new NotFoundException(errorMessage);
```

I think this should be enough as the status code is handled internally

apps/verification/src/verification.service.ts
```diff
@@ -217,7 +217,16 @@ export class VerificationService {
...
-      this.logger.error(`[_getProofPresentationById] - error in get proof presentation by id : ${JSON.stringify(error)}`);
+      this.logger.error(
+        `[_getProofPresentationById] - error in get proof presentation by id : ${JSON.stringify(error)}`
+      );
+      const errorMessage = error?.response?.error?.reason;
+      
+      // Check if error is a "record not found" error
+      if (errorMessage?.includes('record with id') && errorMessage?.includes('not found')) {
+        throw new NotFoundException(errorMessage);
+      }
```
GHkrishna, line 228: I think this is duplication, since we are already checking for a not found exception above, where the function is being called. So I suppose we can either have it here or in `getProofPresentationById`

GHkrishna (changes requested):
Looks good to me just requesting some minor changes

apps/verification/src/verification.service.ts
amrrdev: Thanks for pointing that out; I’ve updated it accordingly

apps/verification/src/verification.service.ts
```diff
@@ -187,17 +187,13 @@ export class VerificationService {
...
 
-      if (errorStack) {
-        throw new RpcException({
-          message: ResponseMessages.verification.error.proofNotFound,
-          statusCode: error?.response?.status,
-          error: errorStack
-        });
-      } else {
-        throw new RpcException(error.response ? error.response : error);
+      if (errorMessage?.includes('record with id') && errorMessage?.includes('not found')) {
```
GHkrishna, line 192: Just a minor change I suppose we can just check for 'not found`. @bhavanakarwade Can you please confirm this once.

bhavanakarwade: Yes @GHkrishna , that’s correct. With this condition, we’re specifically handling the 'not found' case. @amrrdev, could you check if it's sufficient to just check for 'not found' instead of looking for 'record with id' in the error?

amrrdev: I've simplified the error handling to only check for 'not found' as suggested. The updated condition is now more generic and will catch all "not found" scenarios. Let me know if you'd like me to make any other adjustments.

GHkrishna (approved):
Thank for the contribution and for your patience @amrrdev

GHkrishna:
Some of your commits don't seem to be verified, can you please check them once?

amrrdev:
@GHkrishna, I have verified the commit.

GHkrishna:
That's great, thank you for your contribution!

bhavanakarwade:
@amrrdev, Thank you for your patience and contribution.

PR #1200, opened 2025-04-25
feat: Validations for organization and template ledgerId in all issuance methods

bhavanakarwade:
### What?
- Added validations to verify the provided organization and template ledgerId during credential issuance via connection, email, and out-of-band (OOB) methods.


GHkrishna:
This is a temporary fix and will be further fix with the issue #1201

PR #1245, opened 2025-05-13
fix: handle invalid orgID with proper 400 error

### LINKED ISSUE #1220: indy

kirti763:
This PR fixes an issue where a space before an invalid `orgId` in the `GET /v1/webhooks/orgs/webhookurl` API was causing a 500 Internal Server Error.

# Fix

- The `orgId` is now trimmed using the `@Transform` decorator.
- Validation added using `@IsNotEmpty` and `@IsString` to return a 400 Bad Request if `orgId` is missing or invalid.

- Fixes #1220

### DIFF
--- a/apps/api-gateway/src/webhook/dtos/get-webhoook-dto.ts
+++ b/apps/api-gateway/src/webhook/dtos/get-webhoook-dto.ts
@@ -6,11 +6,11 @@ import { trim } from '@credebl/common/cast.helper';
 @ApiExtraModels()
 export class GetWebhookDto {
 
-    @ApiPropertyOptional({example: '2a041d6e-d24c-4ed9-b011-1cfc371a8b8e'})
-    @IsOptional()
+    @ApiProperty({example: '2a041d6e-d24c-4ed9-b011-1cfc371a8b8e'})
     @Transform(({ value }) => trim(value))
+    @IsNotEmpty({ message: 'Please provide the valid orgID' })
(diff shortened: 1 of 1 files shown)


apps/api-gateway/src/webhook/dtos/get-webhoook-dto.ts
```diff
@@ -6,11 +6,11 @@ import { trim } from '@credebl/common/cast.helper';
...
 export class GetWebhookDto {
 
-    @ApiPropertyOptional({example: '2a041d6e-d24c-4ed9-b011-1cfc371a8b8e'})
-    @IsOptional()
+    @ApiProperty({example: '2a041d6e-d24c-4ed9-b011-1cfc371a8b8e'})
     @Transform(({ value }) => trim(value))
+    @IsNotEmpty({ message: 'Please provide the valid orgID' })
     @IsString({ message: 'Organization id must be in string format.' })
-    orgId?: string;
+    orgId: string;
```
GHkrishna, line 13: I think we should still try keeping the value optional as it is. Ideally this should be causing an issue.

platform
```diff
@@ -0,0 +1 @@
+Subproject commit ea4d5f56abc1e7aff64ab7e9d643d2fa19beac6e
```
GHkrishna, line 1: Not sure, what this file is all about. We probably might want to get rid of it

GHkrishna (changes requested):

PR #1247, opened 2025-05-14
requestId-spaces-removed

SEEDART007:
Used trim() function to remove any spaces in front or back of the requestId.


bhavanakarwade:
Hello @SEEDART007,
Thank you for your interest.
We will review your PR, And let you know.

apps/api-gateway/src/issuance/issuance.controller.ts
```diff
@@ -445,7 +445,9 @@ export class IssuanceController {
     @Query() previewFileDetails: PreviewFileDetails,
     @Res() res: Response
   ): Promise<Response> {
-    const { requestId } = query;
+    let { requestId } = query;
+    requestId = requestId.trim();
```
ankita-p17, line 449: There is an interceptor in CREDEBL platform, can you please check if that helps in trim operation.

ankita-p17 (changes requested):

apps/api-gateway/src/issuance/issuance.controller.ts
GHkrishna: We also have a `validation pipe` for `req params` and `query params`

PR #1395, opened 2025-08-20
fix: Save schemaId during connection, email, and bulk issuance

sujitaw:
# What 

- Added fix for connection , email and bulk issuance to save schema id during issuance


apps/issuance/src/issuance.service.ts
```diff
@@ -281,6 +281,18 @@ export class IssuanceService {
 
       if (allSuccessful) {
         finalStatusCode = HttpStatus.CREATED;
+        const context = payload?.credentialData[0]?.credential?.['@context'] as string[];
+
+        if (Array.isArray(context) && context.includes(CommonConstants.W3C_SCHEMA_URL)) {
+          const filterData = context.filter((item) => CommonConstants.W3C_SCHEMA_URL !== item);
+          const [schemaId] = filterData;
+          results.forEach((val) => {
```
shitrerohit, line 289: Please do not provide val as a variable name

PR #1397, opened 2025-08-24
fix: Updated did details in organization by id api

sujitaw:
# What

- Added did created time and last updated on time to API response of organization details.

### DIFF
--- a/apps/organization/interfaces/organization.interface.ts
+++ b/apps/organization/interfaces/organization.interface.ts
@@ -54,7 +54,13 @@ export interface IGetOrgById {
   website: string;
   publicProfile: boolean;
   schema: ISchema[];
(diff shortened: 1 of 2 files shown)


apps/organization/interfaces/organization.interface.ts
```diff
@@ -54,9 +54,15 @@ export interface IGetOrgById {
   website: string;
   publicProfile: boolean;
   schema: ISchema[];
+  didDetails?: IDidCreationData;
   org_agents: IOrgAgents[];
 }
 
+interface IDidCreationData {
```
RinkalBhojani, line 61: I dont think it makes sense to create interface with just the dates and named as IDidCreationData. It may have more details

PR #1402, opened 2025-08-25
Fix: create account with large token size

shitrerohit:
## What?
Modification in account and session table
modification in jwt strategy for handling token expiration

### DIFF
--- a/apps/user/repositories/user.repository.ts
+++ b/apps/user/repositories/user.repository.ts
@@ -681,6 +681,7 @@ export class UserRepository {
       const { sessionToken, userId, expires, refreshToken, accountId, sessionType } = tokenDetails;
       const sessionResponse = await this.prisma.session.create({
         data: {
+          id: tokenDetails.id,
           sessionToken,
           expires,
           userId,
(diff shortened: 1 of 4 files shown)


apps/user/repositories/user.repository.ts
```diff
@@ -682,6 +681,7 @@ export class UserRepository {
       const { sessionToken, userId, expires, refreshToken, accountId, sessionType } = tokenDetails;
       const sessionResponse = await this.prisma.session.create({
         data: {
+          id: tokenDetails.id,
```
tipusinghaw, line 684: We are already destructured `tokenDetails` we can pull out `id` from it.

apps/user/src/user.service.ts
```diff
@@ -475,8 +473,10 @@ export class UserService {
       } else {
         const decryptedPassword = await this.commonService.decryptPassword(password);
         const tokenDetails = await this.generateToken(email.toLowerCase(), decryptedPassword, userData);
-
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        const decodedToken: any = jwt.decode(tokenDetails?.access_token);
```
tipusinghaw, line 477: We can use the type (need to check) jwt.JwtPayload for `decodedToken`

tipusinghaw (approved):

PR #1505, opened 2025-11-04
feat: debug logs

GHkrishna:
## What
- Add debug logs

### DIFF
--- a/apps/oid4vc-verification/src/oid4vc-verification.service.ts
+++ b/apps/oid4vc-verification/src/oid4vc-verification.service.ts
@@ -415,28 +415,4 @@ export class Oid4vpVerificationService extends BaseService {
       throw error;
     }
   }
-
-  // Can have an alternate function, like below
-  /**
-   * async natsCall(pattern: object, payload: object): Promise<{ response: string }> {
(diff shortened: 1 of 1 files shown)


apps/oid4vc-verification/src/oid4vc-verification.service.ts
```diff
@@ -370,4 +421,28 @@ export class Oid4vpVerificationService {
       throw error;
     }
   }
+
+  // Can have an alternate function, like below
```
ajile-in: This commented function may be removed.

apps/oid4vc-verification/src/oid4vc-verification.service.ts
```diff
@@ -28,6 +28,7 @@ import { VerificationSessionQuery } from '../interfaces/oid4vp-verifier.interfac
 import { RequestSignerMethod } from '@credebl/enum/enum';
 
 @Injectable()
+// TODO: Can extend base service, that way the Logger would appear from it
 export class Oid4vpVerificationService {
   private readonly logger = new Logger('IssueCredentialService');
```
RinkalBhojani: Correct this service name in creating logger instance

PR #1556, opened 2026-02-04
refactor: GET all ecosystems API to support orgId as optional query param.

pranalidhanavade:
### What

- Refactored get all ecosystems API to support orgId as optional query param. Added support to fetch ecosystem by ecosystem member, ecosystem lead.
- Removed `IS_ECOSYSTEM_ENABLED` .env variable dependency from user and organization service.

## Summary by CodeRabbit

* **New Features**
  * Optional organization ID filter for ecosystem listings.

* **Bug Fixes / Validation**
  * Stricter UUID/orgId validation and unified bad-request messages across endpoints.
  * Expanded role coverage for ecosystem actions and refined invitation/status handling.
  * Added several user/org-related error messages (e.g., user not found, already accepted).

* **API Changes**
  * Creating an ecosystem now requires name, description, userId and orgId; several ecosystem fields made nullable; user profile email is now required.

* **Chores**
  * Removed legacy static configuration data file.

### DIFF
--- a/apps/ecosystem/src/ecosystem.service.ts
+++ b/apps/ecosystem/src/ecosystem.service.ts
@@ -239,7 +239,11 @@ export class EcosystemService {
 
       const userEmail = user.email;
 
-      const checkUser = await this.ecosystemRepository.getEcosystemInvitationsByEmail(userEmail || '', ecosystemId, orgId);
+      const checkUser = await this.ecosystemRepository.getEcosystemInvitationsByEmail(
+        userEmail || '',
+        ecosystemId,
+        orgId
+      );
 
       if (checkUser && Invitation.REJECTED === checkUser.status && ecosystemId === checkUser.ecosystemId) {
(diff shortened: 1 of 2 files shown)


apps/ecosystem/src/ecosystem.service.ts
```diff
@@ -315,22 +334,54 @@ export class EcosystemService {
...
+        const userId = result.userId || '';
         if (!role) {
           throw new Error('Error fetching ecosystem role');
         }
 
+        if (!result.invitedOrg) {
+          throw new Error(ResponseMessages.ecosystem.error.orgIdMissing);
+        }
+
+        if (!result.userId) {
```
shitrerohit: Please update the sequence: first check whether userId exists, then assign it.

shitrerohit (approved):

apps/ecosystem/src/ecosystem.service.ts
pranalidhanavade: Yes i will update the sequence

Issue #1527, opened 2025-11-26
OpenAPI page redirection


GHkrishna:
## What
- For local development when going to `http://localhost:5000` should ideally redirect to the OpenAPI(Swagger) docs
- Currently it doesn't redirect. Have a look at how we do it here in [agent](https://github.com/credebl/credo-controller)

GHkrishna:
Awesome @TechDev-01 
Assigning it, you start working on it

joaoMiraya:
How is going this issue? I'd like to work on this too
@TechDev-01

PR #1572, opened 2026-02-25
fix: redirect swagger fixed

joaoMiraya:
Redirect http://localhost:5000 -> http://localhost:5000/api when's development mode

[Issue](https://github.com/credebl/platform/issues/1527#event-23023952845)

## Summary by CodeRabbit

* **New Features**
  * Added a root endpoint that redirects to the API documentation when Swagger is enabled.
  * Added graceful fallback behavior when API documentation is unavailable, returning a successful response without a redirect.
  * Enabled Swagger by default in the demo environment, making API documentation readily accessible for demonstration and testing.


RinkalBhojani:
@joaoMiraya - Thanks for your contribution.

I have one suggestion.

Instead of limiting the Swagger UI redirect to development mode only, I think it would be better to make it configurable through some different flag in .env file. That way, users can decide which environments they want to expose the Swagger endpoint in. 

For example, if someone wants to enable it in production for their own use case, they should be able to do so. This keeps the default behavior safe while giving users the flexibility to configure it based on their needs.

joaoMiraya:
> @joaoMiraya - Thanks for your contribution.
> 
> I have one suggestion.
> 
> Instead of limiting the Swagger UI redirect to development mode only, I think it would be better to make it configurable through some different flag in .env file. That way, users can decide which environments they want to expose the Swagger endpoint in.
> 
> For example, if someone wants to enable it in production for their own use case, they should be able to do so. This keeps the default behavior safe while giving users the flexibility to configure it based on their needs.

Its looks great, i gonna make it

ankita-p17:
Hi @joaoMiraya,
Please update .env.demo as well with new variable that you've added.
We are looking forward to approving your contribution & taking it in upcoming release!

joaoMiraya:
> Hi @joaoMiraya, Please update .env.demo as well with new variable that you've added. We are looking forward to approving your contribution & taking it in upcoming release!

Hello, It's done

ankita-p17 (approved):

PR #1573, opened 2026-02-25
Updated vulnerable packages 

DeepakNemad:
- Updated vulnerable packages and Updated all Dockerfiles to use Node.js 24

## Summary by CodeRabbit

* **Infrastructure/DevOps**
  * Upgraded container base to Node 24 Alpine, standardized multi-stage builds, reduced image size, and simplified startup to run compiled Node apps directly.

* **Security**
  * Runtime switched to non-root users, tightened file ownership/permissions, and removed runtime build/migration steps.

* **Dependencies**
  * Large-scale dependency and dev-tooling upgrades (including Prisma and package manager pinning) for more consistent builds.


KambleSahil3 (approved):
LG

PR #1706, opened 2026-08-07
fix(security): harden auth and status-list allocation

### LINKED ISSUE #1705: Security: harden JWT issuer trust, passkey ownership, and status-list allocation
## Summary

Security hardening is needed in three areas:

1. JWT verification derives a JWKS endpoint from the unverified token `iss` claim, allowing an attacker-controlled issuer/JWKS URL to be consulted before issuer validation.
2. Passkey management endpoints do not consistently enforce authentication and ownership, allowing cross-user reads or mutations when a credential ID or email is known.
3. Status-list allocation relies on application-level coordination and lacks database invariants, permitting duplicate allocations under concurrent issuance.

## Proposed fix

- Restrict JWT issuers to an explicit allowlist or the configured Keycloak realm and validate issuer before JWKS lookup.
- Require JWT authentication for passkey endpoints and propagate the authenticated email to enforce device ownership.
- Add transactional allocation locking plus database uniqueness constraints and migration for status-list slots and active issuer allocations.

## Validation

Targeted unit tests cover issuer validation, passkey authorization and ownership, and status-list allocation concurrency.

sign-mark:
## What changed

- Restrict JWT issuer and JWKS resolution to trusted issuers configured through `JWT_TRUSTED_ISSUERS` or the Keycloak realm.
- Require JWT authentication for passkey management and enforce authenticated-user ownership for reads, updates, and deletes.
- Serialize status-list allocation and add database constraints/migration to prevent duplicate allocation under concurrency.

## Why

The previous behavior trusted an unverified JWT issuer while selecting JWKS, allowed passkey operations without consistent caller ownership checks, and relied on application-level coordination for status-list uniqueness.

## Impact

Deployments can optionally configure a comma-separated `JWT_TRUSTED_ISSUERS` allowlist. When omitted, the configured Keycloak domain and realm are used. The included migration adds uniqueness safeguards for issued credential slots and active issuer allocations.

## Validation

- Targeted issuer, passkey, and status-list allocation unit tests: 4 suites, 12 assertions.
- Relevant services compiled successfully.
- Prisma schema, ESLint, and whitespace/diff checks passed.

Fixes #1705

## Summary by CodeRabbit

* **New Features**
  * Added configurable JWT trusted-issuer validation and secure JWKS resolution.
  * Added authenticated ownership checks and actor tracking for passkey operations.
  * Improved passkey credential normalization and status-list allocation safeguards.
  * Added stronger credential-slot uniqueness protections.

* **Bug Fixes**
  * Prevented unauthorized access to other users’ passkeys.
  * Improved handling of concurrent and exhausted status-list allocations.
  * Preserved allocations when credential offers are successfully created.
  * Rejected invalid or unsafe JWT issuer configurations.

* **Tests**
  * Expanded coverage for JWT security, passkey ownership, and status-list allocation.

### DIFF
--- a/apps/api-gateway/src/authz/jwt-issuer.util.ts
+++ b/apps/api-gateway/src/authz/jwt-issuer.util.ts
@@ -38,9 +38,7 @@ export const getTrustedJwtIssuers = (env: NodeJS.ProcessEnv = process.env): stri
   return [...new Set(issuers.map(validateIssuer))];
 };
 
-export const getTrustedJwtIssuerVariants = (trustedIssuers: string[]): string[] => [
-  ...new Set(trustedIssuers.flatMap((issuer) => [issuer, `${issuer}/`]))
-];
+export const getTrustedJwtIssuerVariants = (trustedIssuers: string[]): string[] => [...new Set(trustedIssuers.flatMap((issuer) => [issuer, `${issuer}/`]))];
 
 export const getTrustedJwksUri = (issuer: unknown, trustedIssuers: string[]): string => {
   if ('string' !== typeof issuer) {
--- a/apps/api-gateway/src/authz/jwt-issuer.util.ts
+++ b/apps/api-gateway/src/authz/jwt-issuer.util.ts
@@ -1,6 +1,13 @@
 import { CommonConstants } from '@credebl/common/common.constant';
 
-const trimTrailingSlashes = (value: string): string => value.trim().replace(/\/+$/, '');
+const trimTrailingSlashes = (value: string): string => {
+  const trimmedValue = value.trim();
+  let endIndex = trimmedValue.length;
+  while (0 < endIndex && '/' === trimmedValue.charAt(endIndex - 1)) {
+    endIndex -= 1;
+  }
+  return trimmedValue.slice(0, endIndex);
+};
 
 const validateIssuer = (value: string): string => {
   const issuer = trimTrailingSlashes(value);
@@ -25,11 +32,10 @@ export const getTrustedJwtIssuers = (env: NodeJS.ProcessEnv = process.env): stri
     .map((issuer) => issuer.trim())
     .filter(Boolean);
 
-  const issuers = configuredIssuers?.length
-    ? configuredIssuers
-    : env.KEYCLOAK_DOMAIN && env.KEYCLOAK_REALM
-      ? [`${trimTrailingSlashes(env.KEYCLOAK_DOMAIN)}/realms/${env.KEYCLOAK_REALM.trim()}`]
-      : [];
+  let issuers = [...(configuredIssuers ?? [])];
+  if (0 === issuers.length && env.KEYCLOAK_DOMAIN && env.KEYCLOAK_REALM) {
+    issuers = [`${trimTrailingSlashes(env.KEYCLOAK_DOMAIN)}/realms/${env.KEYCLOAK_REALM.trim()}`];
+  }
 
   if (0 === issuers.length) {
     throw new Error('JWT_TRUSTED_ISSUERS or both KEYCLOAK_DOMAIN and KEYCLOAK_REALM must be configured');
@@ -38,11 +44,13 @@ export const getTrustedJwtIssuers = (env: NodeJS.ProcessEnv = process.env): stri
   return [...new Set(issuers.map(validateIssuer))];
 };
 
-export const getTrustedJwtIssuerVariants = (trustedIssuers: string[]): string[] => [...new Set(trustedIssuers.flatMap((issuer) => [issuer, `${issuer}/`]))];
+export const getTrustedJwtIssuerVariants = (trustedIssuers: string[]): string[] => [
+  ...new Set(trustedIssuers.flatMap((issuer) => [issuer, `${issuer}/`]))
+];
 
 export const getTrustedJwksUri = (issuer: unknown, trustedIssuers: string[]): string => {
   if ('string' !== typeof issuer) {
-    throw new Error('JWT issuer is missing');
+    throw new TypeError('JWT issuer is missing');
   }
 
   const normalizedIssuer = trimTrailingSlashes(issuer);
--- a/apps/api-gateway/src/authz/jwt-issuer.util.ts
+++ b/apps/api-gateway/src/authz/jwt-issuer.util.ts
@@ -4,7 +4,12 @@ const trimTrailingSlashes = (value: string): string => value.trim().replace(/\/+
 
 const validateIssuer = (value: string): string => {
   const issuer = trimTrailingSlashes(value);
-  const parsed = new URL(issuer);
+  let parsed: URL;
+  try {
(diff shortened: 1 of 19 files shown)


sign-mark:
@coderabbitai review

sign-mark:
@coderabbitai review

sign-mark:
@coderabbitai review

sign-mark:
@coderabbitai review

apps/user/src/fido/fido.service.ts
```diff
@@ -1,246 +1,283 @@
...
-            } else {
-                return 'Device name has not been changed.';
-            }
-        } catch (error) {
-            this.logger.error(`Error in delete user device :::${error}`);
-            throw new RpcException(error);
-        }
+  private async assertDeviceOwnership(credentialId: string, actorEmail: string): Promise<PasskeyDeviceOwnership> {
+    const actor = await this.fidoUserRepository.checkFidoUserExist(actorEmail.toLowerCase());
+    const device = await this.userDevicesRepository.checkUserDeviceByCredentialId(credentialId);
```
ajile-in, line 269: [blocker] Think this locks owners out of freshly registered devices. `createMultiDevice` never sets the `credentialId` column — it only stashes the credential inside the `devices` JSON (padding stripped). The column gets populated later by `addCredentialIdAndNameById`, which runs inside `updateUser`… which is gated by this very check. So the first rename/update on a brand-new device finds no row and throws Forbidden, even for the rightful owner — bit of a chicken-and-egg.

Suggestion: set `credentialId` at creation time in `verifyRegistration` (from `newDevice.credentialID`), or fall back to matching `devices->>'credentialID'` here. The specs miss this because their fixtures pre-populate the column.

apps/user/src/fido/fido.service.ts
```diff
@@ -1,246 +1,283 @@
...
+
+    if (!device || device.userId !== actor.id || device.deletedAt) {
```
ajile-in: [minor] If the email doesn't match a user row, `checkFidoUserExist` (`findFirstOrThrow`) blows up before we even get here — and if it ever returned undefined, this line would TypeError. Either way the caller sees a generic 500 instead of a clean 403/404. An explicit null-check on `actor` would make failures much easier to reason about.

ajile-in (commented):
Thanks for tackling this — all three fixes are aimed at real issues. The issuer allowlist closes a genuine bypass (the old strategy built the JWKS URL straight from the token's `iss` and swallowed JWKS errors), the passkey ownership checks close an IDOR on delete/rename, and the advisory lock + unique index should hold up across replicas. Tests pass locally (4 suites / 19 tests), lint clean.

One thing below looks like it'll break first-time device setup — worth sorting before merge. Three small nits for the road:

- `[nit]` `new URL()` in `validateIssuer` throws a bare TypeError instead of your friendly "Invalid trusted JWT issuer" message
- `[nit]` trailing slashes are trimmed for the JWKS lookup but jsonwebtoken's `issuer` check compares raw, so a token with a trailing-slash `iss` passes one gate and fails the other
- `[nit]` if `allocatedCount` ever drifts from the bitmap, the catch path surfaces a 500 ("Status list bitmap is full") instead of rolling a fresh list

apps/api-gateway/src/fido/fido.controller.ts
```diff
@@ -222,7 +246,8 @@ export class FidoController {
   ): Promise<Response> {
     const verifyRegistration = await this.fidoService.updateFidoUser(
       updateFidoUserDetailsDto,
-      decodeURIComponent(credentialId)
+      decodeURIComponent(credentialId),
```
ajile-in, line 249: [minor] Encoding gotcha: the JSON copy of the credential is padding-stripped base64url (see the `.replace(/=*$/, '')` in verifyRegistration), while this comes straight off the URL. A client sending padding or standard base64 (`+`/`/`) can silently miss both the ownership lookup and the JSON match downstream. Normalizing both sides before comparing would remove the footgun.

apps/api-gateway/src/authz/jwt-issuer.util.ts
```diff
@@ -0,0 +1,47 @@
...
+    .filter(Boolean);
+
+  const issuers = configuredIssuers?.length
+    ? configuredIssuers
+    : env.KEYCLOAK_DOMAIN && env.KEYCLOAK_REALM
+      ? [`${trimTrailingSlashes(env.KEYCLOAK_DOMAIN)}/realms/${env.KEYCLOAK_REALM.trim()}`]
+      : [];
+
+  if (0 === issuers.length) {
+    throw new Error('JWT_TRUSTED_ISSUERS or both KEYCLOAK_DOMAIN and KEYCLOAK_REALM must be configured');
```
ajile-in, line 41: [note] Fail-closed is the right call here — just flagging the ops impact: any deployment with `KEYCLOAK_DOMAIN` set but no `KEYCLOAK_REALM` will now refuse to boot. Worth a line in the release notes so nobody's surprised mid-deploy.

libs/prisma-service/prisma/migrations/20260806090000_harden_status_list_allocation/migration.sql
```diff
@@ -0,0 +1,42 @@
...
+  END IF;
+END $$;
+
+-- Prevent two service replicas from persisting the same status-list slot.
+CREATE UNIQUE INDEX "issued_oid4vc_credentials_listId_index_key"
+ON "issued_oid4vc_credentials"("listId", "index");
+
+-- There must be at most one active list for a tenant/issuer pair. Prisma does
+-- not currently model partial indexes, so this invariant lives in SQL.
+CREATE UNIQUE INDEX "status_list_allocation_one_active_per_issuer_key"
```
ajile-in, line 39: [nit] This partial index only exists in SQL since Prisma can't model it — totally fine, just be aware that beyond drift warnings, a future `prisma migrate dev` diffs the shadow DB against schema.prisma and could generate a `DROP INDEX` for it. Worth keeping an eye on if CI runs drift checks.

sign-mark:
@ajile-in Thanks again for the thorough review. I’ve addressed the blocker and follow-up nits, added regression coverage, and resolved the related threads. All checks are green—could you please take another look when you have a moment?

ajile-in:
@tipusinghaw , @ankita-p17 - Any thoughts on this PR?

sign-mark:
@tipusinghaw @ankita-p17 All reported blockers and follow-up comments have been addressed, the relevant threads were resolved, and DCO/SonarCloud/CodeRabbit are green. Could one of you please provide the remaining approval when convenient?

ankita-p17:
> @tipusinghaw @ankita-p17 All reported blockers and follow-up comments have been addressed, the relevant threads were resolved, and DCO/SonarCloud/CodeRabbit are green. Could one of you please provide the remaining approval when convenient?

Hi @sign-mark, we are testing e2e with these changes & will update the PR soon.

